PR #38, opened 2021-05-18
Add SqLite support to SAF.Storage

MarkusFriesen:
Add SqLite as another Storage Infrastructure option.


src/Storage/SAF.Storage.SqLite/Storage.cs
```diff
@@ -0,0 +1,123 @@
...
+
+using SAF.Common;
+using System;
+using System.Data.SQLite;
+
+namespace SAF.Storage.SqLite
+{
+    public class Storage : IStorageInfrastructure, IDisposable
+    {
+        readonly SQLiteConnection _connection;
```
berndbr, line 13: Pls use explicit access modifier "private"

src/Storage/SAF.Storage.SqLite/Storage.cs
```diff
@@ -0,0 +1,123 @@
...
+        readonly string bytesValue = "bytesValue";
```
berndbr, line 14: Could be "private const string"

PR #43, opened 2021-07-29
Replace registry alive

boeschen67:


src/Communication/SAF.Communication.PubSub.Cde/SubscriptionRegistry.cs
```diff
@@ -276,6 +273,13 @@ private void HandleSubscriberAlive(TSM message)
                 {
                     _syncSubscribers.ExitWriteLock();
                 }
+                if (string.IsNullOrWhiteSpace(message.PLS))
+                {
+                    var tsmAlive = new TSM(Engines.PubSub, MessageToken.RegistryAlive, _registryIdentity);
```
berndbr, line 278: We should answer to `message.ENG` instead of `Engines.PubSub` to catch the case of "remote subscribers" too.

src/Communication/SAF.Communication.PubSub.Cde/SubscriptionRegistry.cs
```diff
@@ -276,6 +273,13 @@ private void HandleSubscriberAlive(TSM message)
...
+                    tsmAlive.SetToServiceOnly(true);
```
berndbr, line 279: When answering to `message.ENG`, we should not send to service only (`SetToServiceOnly(true)`). Remove that line to send the answer TSMs to nodes (e.g. browser instances) too.

berndbr (approved):

PR #83, opened 2024-12-16
Add retry for file chunks

NicoKno:
Added simple retry if not single chunk was not delivered, default is no retry, attempts configurable


src/Toolbox/SAF.Toolbox/ServiceCollectionExtensions.cs
```diff
@@ -63,9 +63,18 @@ public static IServiceCollection AddRequestClient(this IServiceCollection servic
         return services;
     }
 
-    public static IServiceCollection AddFileSender(this IServiceCollection services)
+    public static IServiceCollection AddFileSender(this IServiceCollection services, IConfiguration? hostConfig = null)
```
berndbr, line 66: This is a change in the public API that breaks . To avoid major version update it might be better to add a additional method `AddFileSender` with the new parameter.

berndbr (changes requested):

src/Toolbox/SAF.Toolbox/ServiceCollectionExtensions.cs
NicoKno: I don't know if this is better, but I changed it. I think the default value would have preserved the compatibility.

berndbr (approved):

PR #106, opened 2026-07-13
Migrate SAF to the new plugin system

berndbr:
This PR migrates SAF from the legacy plugin-loading approach to the new SAF.PluginSystem.* model.

**Includes**
•	new host integration via AddSafHost()
•	plugin-based loading for SAF core modules
•	messaging refactoring to the new plugin/runtime model
•	migrated test runners and sample plugins
•	removal of old plugin-system/dev-tooling code
•	updated tests and documentation

**Notes**
AddSafHost() now loads built-in SAF plugins and registers default SAF contract assemblies, while still allowing applications to extend configuration.


src/PluginSystem/SAF.PluginSystem.Hosting/ServiceProviderExtensions.cs
```diff
@@ -0,0 +1,63 @@
...
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+using System.Collections;
+using System.IO.Abstractions;
+
+internal static class ServiceProviderExtensions
+{
+    public static IServiceProvider RedirectCommonServices(this IServiceProvider sp, IServiceCollection services)
+    {
+        services.AddSingleton(_ => sp.GetRequiredService<ILoggerFactory>());
```
MarkusFriesen: **High — plugin containers dispose the shared host `ILoggerFactory`.** These host singletons are redirected via factory delegates (`_ => sp.GetRequiredService<…>()`). MEDI captures the `IDisposable` result of a *factory* registration and disposes it when the **plugin** container is disposed — so each plugin container takes ownership of and disposes the host's shared `ILoggerFactory` (once per container, before the host disposes its own). After the first `PluginServicesContainer` teardown the shared factory is dead, breaking shutdown logging / telemetry flush and risking `ObjectDisposedException` from later `CreateLogger`. Note the asymmetry with `HostServiceForwarder.Forward`, which uses the *instance* overload (not captured).

Fix: keep resolution lazy but wrap the factory in a non-owning adapter whose `Dispose()` is a no-op, so the plugin container disposes only the wrapper. (Naively switching to the instance overload instead forces eager resolution and breaks the container's lazy contract — it regresses 19 existing tests.)

src/PluginSystem/SAF.PluginSystem.Hosting/PublicPluginServiceFactory.cs
```diff
@@ -0,0 +1,96 @@
...
+        {
+            return _cachedInstance ??= ResolveInternal(serviceProvider);
+        }
+    }
+
+    private object? ResolveKeyedInternal(IServiceProvider serviceProvider, object? serviceKey)
+    {
+        if (_ownedDescriptor.KeyedImplementationInstance is not null)
+        {
+            return _ownedDescriptor.ImplementationInstance;
```
MarkusFriesen: **Keyed-instance public service throws at resolve.** When `KeyedImplementationInstance is not null` this returns the **non-keyed** `ImplementationInstance` accessor, whose getter throws `InvalidOperationException` on a keyed `ServiceDescriptor` (keyed-descriptor misuse). Any public contract registered as `AddKeyedSingleton<T>(key, instance)` — a supported cross-plugin pattern — crashes here.

Fix: `return _ownedDescriptor.KeyedImplementationInstance;`

src/PluginSystem/SAF.PluginSystem.Hosting/PublicPluginServiceFactory.cs
```diff
@@ -0,0 +1,96 @@
...
+
+    public object? Resolve(IServiceProvider serviceProvider)
+    {
+        if(_isKeyedService)
+        {
+            return _cachedInstance ??= ResolveKeyedInternal(serviceProvider, _serviceKey);
+        }
+        else
+        {
+            return _cachedInstance ??= ResolveInternal(serviceProvider);
```
MarkusFriesen: **Unsynchronized lazy cache can break public-singleton identity.** `_cachedInstance ??= …` is not atomic; two threads resolving the same public singleton for the first time concurrently can both observe `null` and construct two instances, so consumers across plugins can receive different objects. Reachable because `InProcessMessaging.Publish` dispatches handlers via `Task.Run`. Guard the `??=` with a lock.

src/PluginSystem/SAF.PluginSystem.Hosting/PluginManifestLoader.cs
```diff
@@ -0,0 +1,28 @@
...
+using Contracts;
+using System.Reflection;
+
+internal sealed class PluginManifestLoader : IPluginManifestLoader
+{
+    public IPluginManifest? LoadPluginManifest(Assembly assembly)
+    {
+        ArgumentNullException.ThrowIfNull(assembly);
+
+        foreach (var type in assembly.GetTypes())
```
MarkusFriesen: **One bad DLL aborts discovery of all plugins.** `assembly.GetTypes()` throws `ReflectionTypeLoadException` if any type cannot be loaded (missing/incompatible transitive dependency), and the caller (`PluginAssemblyFolderContainer.LoadManifests`, materialized eagerly under the cache lock) has no per-assembly guard — nor does `LoadFromAssemblyPath` (`BadImageFormatException` for a native DLL matched by `*.dll`). Net: a single unloadable assembly in the plugin folder silently kills discovery of **every** plugin.

Fix: catch `ReflectionTypeLoadException` here (fall back to `ex.Types`, skipping nulls), and wrap the per-assembly load in `LoadManifests` in a try/catch that logs and continues.

src/PluginSystem/SAF.PluginSystem.Hosting/ServicePluginHost.cs
```diff
@@ -0,0 +1,110 @@
...
+            catch (Exception ex)
+            {
+                logger.LogError(ex, "Failed to stop service plug-in.");
+            }
+        }
+
+        logger.LogInformation("Service plugins stopped.");
+    }
+
+    public async Task StartingAsync(CancellationToken cancellationToken)
```
MarkusFriesen, line 60: **Lifecycle callbacks lack the exception isolation `StartAsync`/`StopAsync` have.** `StartingAsync`/`StartedAsync`/`StoppingAsync`/`StoppedAsync` `await` each plugin with no try/catch, so one plugin throwing — especially during `StoppingAsync`/`StoppedAsync` on shutdown — skips the callback for all remaining plugins and their cleanup/flush is lost. Wrap each call in try/catch + `LogError`, matching lines 24-31 / 47-54.

MarkusFriesen (commented):
## Deep-dive review — migration to the new plugin system

CI is green and the architecture is sound. This is a thorough pass over the new `SAF.PluginSystem.Hosting` DI/isolation core and the reworked messaging providers. Inline comments below cover the **1 high + 9 medium** issues; lower-severity items are grouped here. I have verified fixes (build + **76/76** `SAF.PluginSystem.Hosting` tests green) for the five core plugin-system issues (shared `ILoggerFactory` disposal, keyed-instance throw, `GetTypes` abort, lifecycle isolation, cache race).

### Lower-severity findings
- **NATS `Subscribe` blocks ~30s then throws when the broker is down** — `SAF.Messaging.NATS/Messaging.cs:178`: `WaitUntil(() => isSynchronized)` times out (30s) and `.Wait()` throws `AggregateException(TimeoutException)`, uncaught by the `NullReferenceException`/`ObjectDisposedException` handlers; the background subscribe task is left faulted/unobserved and keeps retrying (`MaxReconnectRetry=-1`). (Line is outside the diff, so not inline.)
- **Transient storage leaks a connection per resolve** — NATS `ServiceCollectionExtensions.cs:202` and Redis `ServiceCollectionExtensions.cs:146`: `AddTransient<IStorageInfrastructure>` opens a fresh `NatsClient`/`ConnectionMultiplexer` on every resolution, retained until container disposal. LiteDb/SQLite/Cde use `AddSingleton`; consider matching.
- **`MessageHandlerResolver` materializes every handler per dispatch** — `SAF.Messaging.Extensions/MessageHandlerResolver.cs:25`: `GetServices<IMessageHandler>()` constructs all handlers on each message to pick one by `GetType()`; leaks transient `IDisposable` handlers via the root provider.
- **NATS `Messaging.Dispose` does not cancel subscriptions / dispose the `CancellationTokenSource`s** — `SAF.Messaging.NATS/Messaging.cs:145`.
- **Routing self-reference → `StackOverflow`** — `SAF.Messaging.Routing/ServiceCollectionExtensions.cs:50`: a route whose `Messaging.Key == "Routing"` recurses unbounded.
- **Empty `MessageRouting` config + `PrimaryKey=Routing` silently drops all messages** — `SAF.Messaging.Routing/PluginManifest.cs:21`.
- **`Recursive` + a `**` glob double-counts assemblies** — `PluginAssemblyFolderContainer.cs:70` (duplicate plugin loads).
- **NATS `ConnectionTimeout` is dead config** — `SAF.Messaging.NATS/ServiceCollectionExtensions.cs:98` is never mapped to `NatsOpts`/`ConnectAsync`.
- **`InProcessMessaging.Dispose` never disposes its two `ReaderWriterLockSlim`** — `InProcessMessaging.cs:206`.
- **`AddPluginSystem` overrides an app-provided `IFileSystem`** — `HostApplicationBuilderExtensions.cs:22` (use `TryAddSingleton`).
- **`PluginAssemblyLoadContext` re-enumerates the base dir on every `Load`** when nothing resolves — `PluginAssemblyLoadContext.cs:78` (init guard keys on `Count > 0`).
- Also worth a look: `PluginServiceProvider.GetService<T>()` uses `SingleOrDefault()` and **throws** if two plugins expose the same public contract type.

_Checked and dismissed (no action needed): `CollectPublicServices` "perf" (runs once over a ~2-item set); dispatcher "double-dispatch" (handling count is invariant to the cache race); NATS `PublishAsync` discarded `ValueTask` (pre-existing, unchanged by this PR)._

src/PluginSystem/SAF.PluginSystem.Hosting/PluginServicesLocatorBuilder.cs
```diff
@@ -0,0 +1,139 @@
...
+                var pluginServiceLocator = new ImportedServiceLocator(collection, publicServiceDescriptor);
+
+                if (publicServiceDescriptor.IsKeyedService)
+                {
+                    AddKeyedService(_pluginServices.ServiceCollection,
+                        (sp, key) => pluginServiceLocator.GetKeyedService(publicServiceDescriptor.ServiceType, key)!, publicServiceDescriptor);
+                }
+                else
+                {
+                    AddService(_pluginServices.ServiceCollection,
```
MarkusFriesen, line 91: **Shared public `IDisposable` singleton is disposed once per importing container.** A public singleton owned by plugin A is re-registered into every importing plugin's collection (and the public-only collection) as that container's own singleton (via `CreateInstanceByFactory` / the imported factory). Each of those containers then disposes the **same** instance on teardown → double/N-fold `Dispose()`. For an owner-only-disposes model, imported/public registrations should resolve the instance without the importing container taking disposal ownership. (Same family as the `ILoggerFactory` cross-ownership issue.)

src/PluginSystem/SAF.PluginSystem.Hosting/PluginSystemHostBuilderExtensions.cs
```diff
@@ -0,0 +1,42 @@
...
+        return hostBuilder.AddPluginAssemblyFolderContainer(_ => { });
+    }
+
+    public static IPluginSystemHostBuilder AddPluginAssemblyFolderContainer(this IPluginSystemHostBuilder hostBuilder, Action<PluginAssemblyFolderSearchOptions> configureSearch)
+    {
+        ArgumentNullException.ThrowIfNull(hostBuilder);
+
+        var uniqueOptionsKey = $"{nameof(PluginAssemblyFolderSearchOptions)}_{Guid.NewGuid():N}";
+        hostBuilder.Services.Configure(uniqueOptionsKey, configureSearch);
+        hostBuilder.Services.AddTransient<IPluginAssemblyContainer, PluginAssemblyFolderContainer>(sp =>
```
MarkusFriesen: **`IPluginAssemblyContainer` registered `Transient` → duplicate, non-collectible assembly loads.** Two singletons consume `IEnumerable<IPluginAssemblyContainer>`: `PluginServicesContainer` and `ServiceHostDiagnostics`. Because the registration is transient, each gets its **own** `PluginAssemblyFolderContainer` instances with empty per-instance caches, so `ServiceHostDiagnostics` triggers a **second** folder scan that, for every plugin assembly outside `AppContext.BaseDirectory`, creates a fresh **non-collectible** `PluginAssemblyLoadContext` and re-loads the assembly (never unloaded) purely to read reflection metadata. Fix: register as `AddSingleton` (the container owns a cache + ALC lifetime), or have diagnostics read the already-materialized manifests via `IPluginServicesContainer`.

src/Messaging/SAF.Messaging.Cde/ServiceCollectionExtensions.cs
```diff
@@ -50,24 +49,20 @@ public static IServiceCollection AddCdeInfrastructure(this IServiceCollection co
...
-                    sp.GetRequiredService<ISubscriber>(),
-                    null,
-                    new CdeMessagingConfiguration(cfg))))
-            .AddTransient(sp => sp.GetRequiredService<Func<MessagingConfiguration, ICdeMessagingInfrastructure>>().Invoke(config));
+    private static Messaging CreateMessagingInfrastructure(IServiceProvider serviceProvider, MessagingConfiguration config)
+        => new Messaging(serviceProvider.GetService<ILogger<Messaging>>(),
+            ResolveMessageDispatcher(serviceProvider),
+            serviceProvider.GetRequiredService<IPublisher>(),
+            serviceProvider.GetRequiredService<ISubscriber>(),
+            null,
```
MarkusFriesen: **`traceAction` is silently dropped for CDE (regression).** `AddCdeMessagingInfrastructure(traceAction)` / `AddCdeInfrastructure(traceAction)` accept the callback, but the keyed factory closure and `CreateMessagingInfrastructure` never use it — `new Messaging(...)` is passed `null` here, so `Messaging.Publish`'s `_traceAction?.Invoke(message)` never fires. InProcess/NATS/Redis all wire it through, and pre-PR CDE did too. Fix: thread `traceAction` into the closure and `CreateMessagingInfrastructure` → `new Messaging(..., traceAction)`.

src/Messaging/SAF.Messaging.Routing/ServiceCollectionExtensions.cs
```diff
@@ -33,86 +29,33 @@ public static IServiceCollection AddRoutingMessagingInfrastructure(this IService
...
         return config.Routings
             .Select(r =>
             {
-                var serviceType = serviceCollection.FirstOrDefault(sd => sd.ServiceType.Name == r.Messaging.Type)?.ServiceType;
-                if(serviceType == null)
-                    throw new FileNotFoundException($"Messaging DLL not installed for messaging type {r.Messaging.Type}");
-
-                var routing = new MessageRouting(BuildMessagingInfrastructure(serviceProvider, serviceType, r.Messaging))
+                var factory = serviceProvider.GetRequiredKeyedService<IMessagingInfrastructureFactory>(r.Messaging.Key);
+                var routing = new MessageRouting(factory.Create(r.Messaging))
```
MarkusFriesen, line 57: **Routing leaks the backend connections it creates.** `BuildMessageRouting` builds each backend via `factory.Create(r.Messaging)`, which runs `new Messaging(...)` inside the delegate — so these `IDisposable` backends (NATS `NatsClient`, Redis `ConnectionMultiplexer`) are **not** tracked by the container. They are held inside the routing `Messaging`, which is **not** `IDisposable` and has no cleanup. Contrast the non-routing primary path (e.g. `PrimaryKey=Nats`), where the Runtime factory returns the backend to DI and it *is* disposed. Fix: make the routing wrapper `IDisposable` and dispose its backends, or resolve backends through the container so it owns disposal.

src/Hosting/SAF.Hosting/ServiceCollectionExtensions.cs
```diff
@@ -1,66 +1,53 @@
...
-
-        return builder;
+        ArgumentNullException.ThrowIfNull(services);
+        ArgumentNullException.ThrowIfNull(configure);
+
+        services.Configure(configure);
+        services.AddSingleton<IServiceHostInfo>(sp =>
+        {
+            var options = sp.GetRequiredService<IOptions<ServiceHostOptions>>().Value;
+            return new ServiceHostInfo(options, () => GetOrInitializeHostId(sp.GetService<IStorageInfrastructure>()));
```
MarkusFriesen: **Host-ID persistence is dead code → new host id every start.** `GetOrInitializeHostId(sp.GetService<IStorageInfrastructure>())` resolves storage from the host **root** provider, but in the new architecture `IStorageInfrastructure` is only ever registered inside **plugin** containers (SQLite/LiteDb/messaging manifests) — never in the host root. So the resolve is always `null`, the read/write-back path is a no-op, and a fresh GUID is minted each process start (stable only within a run via `ServiceHostInfo`'s memoization). Regression from the old shared-container `ServiceHost`. Fix: source storage from a shared/plugin registry, or forward the host id through plugin storage explicitly.

MarkusFriesen (approved):

PR #107, opened 2026-07-27
Introduce reload of plugin containers

berndbr:
Allow to reload the plug-in DI containers to reflect configuration changes.


src/PluginSystem/SAF.PluginSystem.Hosting/PluginServicesContainer.cs
```diff
@@ -36,6 +36,33 @@ public IServiceProvider GetPublicServices()
...
+
+            logger.LogInformation("Reinitializing plug-in service providers.");
+
+            providersToDispose = SnapshotProviders();
+
+            _pluginServiceCollections = [];
+            _publicServicesOnlyCollection = new PluginServiceCollection(new ServiceCollection(), []);
+            _initialized = true;
+
+            BuildProviders();
```
berndbr: **Reload has no rollback: one throwing manifest permanently breaks the container and leaks the previous generation.**

`BuildProviders()` runs *after* the state has already been swapped (`_pluginServiceCollections = []`, `_publicServicesOnlyCollection = new(...)`, `_initialized = true`). If any `IPluginManifest.ConfigureServices` throws â€” e.g. the changed settings contain a bad value and a manifest does `ArgumentException`/`InvalidOperationException` â€” the exception escapes with the container in a wedged state and `providersToDispose` (the entire previous generation) is discarded before the dispose loop is reached.

I ran the real type to confirm:

```
ReinitializeAsync threw: InvalidOperationException: configure boom
OLD gen1 provider still usable -> NOT disposed -> leaked
GetPublicServices() after failed reinit -> NULL
GetPluginServices() after failed reinit -> count=1, nulls=1
ServicePluginHost-style resolve threw ArgumentNullException: Value cannot be null. (Parameter 'provider')
```

Consequences:
- `GetPublicServices()` returns `null` even though the declared return type is non-nullable (`<Nullable>enable</Nullable>`), so `PluginServiceProvider.GetServices<T>()` throws for every host caller.
- `GetPluginServices()` yields a list containing `null`, so `ServicePluginHost.GetServicePlugins()` and `PluginSystemController.GetServicePlugins()` throw `ArgumentNullException`.
- `_initialized` stays `true`, so nothing ever retries the build â€” unrecoverable without a process restart.
- The old generation is never disposed, not even by `DisposeAsync`, because `SnapshotProviders()`'s `.OfType<IServiceProvider>()` filters out the `null` `ServiceProvider`s of the half-built new generation.

Build into local variables first and only assign the fields once every provider is built, restoring the old generation on failure.

src/PluginSystem/SAF.PluginSystem.Hosting/PluginServicesContainer.cs
```diff
@@ -21,8 +21,8 @@ public sealed class PluginServicesContainer(
     private bool _initialized = false;
     private bool _disposed = false;
 
-    private readonly List<PluginServiceCollection> _pluginServiceCollections = [];
-    private readonly PluginServiceCollection _publicServicesOnlyCollection = new(new ServiceCollection(), []);
+    private List<PluginServiceCollection> _pluginServiceCollections = [];
```
berndbr, line 31: **Making these fields mutable introduces a data race: `GetPluginServices()` / `GetPublicServices()` read them outside the lock.**

```csharp
public IServiceProvider GetPublicServices()
{
    InitializeServiceProviders();                       // takes AND RELEASES the lock
    return _publicServicesOnlyCollection.ServiceProvider!;   // read with NO lock held
}
```

While these were `readonly` and populated exactly once, the unsynchronized read was benign. Now `ReinitializeAsync` replaces both fields, so a reader that passes `InitializeServiceProviders()` and is then preempted observes the freshly-created generation whose `ServiceProvider`s are still `null`.

Measured against the real type (64 reader threads, 208,321 `ReinitializeAsync` cycles):

```
GetPublicServices -> null  : 1291
GetPluginServices has null : 676
```

Both members declare non-nullable returns, so callers (`PluginServiceProvider`, `ServicePluginHost`, `PluginSystemController`) get a `NullReferenceException` / `ArgumentNullException` instead. Return the snapshot from *inside* the lock rather than re-reading the fields after releasing it.

src/PluginSystem/SAF.PluginSystem.Hosting/PluginSystemController.cs
```diff
@@ -0,0 +1,59 @@
...
+using Contracts;
+using Microsoft.Extensions.DependencyInjection;
+using Microsoft.Extensions.Logging;
+
+/// <inheritdoc />
+internal sealed class PluginSystemController(
+    ILogger<PluginSystemController> logger,
+    IPluginServicesContainer pluginServicesContainer) : IPluginSystemController
+{
+    public async Task ReloadAsync(CancellationToken cancellationToken = default)
```
berndbr, line 24: **`ReloadAsync` is not serialized, so two overlapping reloads double-start plugins.**

Nothing guards against concurrent entry. With two in-flight reloads the interleaving `A.stop â†’ A.reinit â†’ B.stop â†’ B.reinit â†’ A.start â†’ B.start` makes both `A` and `B` resolve *the same* (newest) generation and call `StartAsync` on it, with no `StopAsync` in between.

Measured with the same stop/reinit/start sequence over 4,000 iterations of two concurrent reloads:

```
live plugin StartAsync called >1x  : 19
```

This is not a theoretical trigger â€” `docs/plugin-system.md:428` explicitly recommends wiring this to `IOptionsMonitor.OnChange` or a file watcher, and both fire twice for a single file save (editors and deploy tooling write non-atomically). A plugin that binds a socket, starts a timer, or subscribes to a broker in `StartAsync` will double-bind or throw.

Add a `SemaphoreSlim(1,1)` around the whole `ReloadAsync` body (and consider coalescing rapid triggers).

src/PluginSystem/SAF.PluginSystem.Hosting/PluginSystemController.cs
```diff
@@ -0,0 +1,59 @@
...
+    {
+        logger.LogInformation("Reloading plugin system.");
+
+        await StopServicePluginsAsync(cancellationToken).ConfigureAwait(false);
+        await pluginServicesContainer.ReinitializeAsync(cancellationToken).ConfigureAwait(false);
```
berndbr: **If `ReinitializeAsync` throws or the token cancels, every plugin is left stopped â€” silently and permanently.**

`ReloadAsync` has no `try`/`catch`, no rollback and no restart path:

```csharp
await StopServicePluginsAsync(cancellationToken);
await pluginServicesContainer.ReinitializeAsync(cancellationToken);  // throws ->
await StartServicePluginsAsync(cancellationToken);                   // never runs
```

So when the rebuild fails (see the no-rollback issue in `PluginServicesContainer.BuildProviders`) or the caller cancels between the stop and the rebuild, the plugin system ends up fully stopped, `"Plugin system reloaded."` is never logged, *no* error is logged at the controller level, and the exception surfaces in whatever callback triggered the reload â€” typically an `IOptionsMonitor.OnChange` handler, where it is swallowed or crashes a thread-pool thread.

Combined with the wedged-container issue, a single bad configuration value takes the whole host down to zero running plugins with no way back short of a restart.

Also note `GetServicePlugins()` on line 29 is called *outside* the per-plugin `try`, so an exception from resolving the plugin list propagates out of `ReloadAsync` too.

src/PluginSystem/SAF.PluginSystem.Hosting/PluginServicesContainer.cs
```diff
@@ -36,6 +36,33 @@ public IServiceProvider GetPublicServices()
...
+            _pluginServiceCollections = [];
+            _publicServicesOnlyCollection = new PluginServiceCollection(new ServiceCollection(), []);
+            _initialized = true;
+
+            BuildProviders();
+        }
+
+        foreach (IServiceProvider provider in providersToDispose)
+        {
+            await DisposeProviderAsync(provider).ConfigureAwait(false);
```
berndbr: **Disposing the previous generation also disposes host-owned singletons.**

`ServiceProviderExtensions.RedirectCommonServices` injects host services into every plugin collection using **factories**, not instances:

```csharp
services.AddSingleton(_ => sp.GetRequiredService<IPluginServiceProvider>());
services.AddSingleton(_ => sp.GetRequiredService<IPluginSystemHostEnvironment>());
services.AddSingleton(_ => sp.GetRequiredService<IFileSystem>());
```

Microsoft DI captures and disposes whatever a factory returns, so each plugin provider believes it owns these host singletons. Until now that was harmless because plugin providers were only disposed at process shutdown. `ReinitializeAsync` disposes them mid-life, so the host's own singletons get disposed while the host keeps running.

Verified by building a plugin generation over a host provider and reloading:

```
host IPluginSystemHostEnvironment disposed : True
```

The shipped implementations happen not to be `IDisposable` today, but all three are public contract interfaces a host can implement, and `IFileSystem` is a `TryAddSingleton` so a host can substitute a disposable one. Any service a custom `IHostServiceForwarder` forwards via a factory has the same problem. `ILoggerFactory` is already protected by `NonOwningLoggerFactory` â€” the hazard was known but not generalized. Wrap all redirected host services non-owningly (or register them as instances, which DI does not dispose).

src/PluginSystem/SAF.PluginSystem.Hosting/PluginSystemController.cs
```diff
@@ -0,0 +1,59 @@
...
+        await StartServicePluginsAsync(cancellationToken).ConfigureAwait(false);
+
+        logger.LogInformation("Plugin system reloaded.");
+    }
+
+    private async Task StartServicePluginsAsync(CancellationToken cancellationToken)
+    {
+        foreach (IServicePlugin servicePlugin in GetServicePlugins())
```
berndbr: **Reload skips every `ILifecycleServicePlugin` phase that `ServicePluginHost` honours.**

`ServicePluginHost` implements `IHostedLifecycleService` and calls `StartingAsync` / `StartedAsync` / `StoppingAsync` / `StoppedAsync` on plugins implementing `ILifecycleServicePlugin`. `ReloadAsync` only calls `StopAsync` and `StartAsync`.

A plugin that does its cross-plugin wiring in `StartedAsync` (the documented purpose: "Called after `StartAsync` has completed to perform post-start operations") or releases resources in `StoppedAsync` therefore ends up half-initialized after every reload, while a host restart works correctly â€” a difference that is very hard to debug from the plugin side.

The copy also loses `ServicePluginHost`'s `CancellationTokenSource.CreateLinkedTokenSource(cancellationToken)` wrapper.

src/PluginSystem/SAF.PluginSystem.Hosting/PluginSystemHostContext.cs
```diff
@@ -46,12 +46,12 @@ private static IConfiguration BuildPluginConfiguration(
                     options.PluginSettingsFilePath, settingsFilePath);
             }
 
-            builder.AddJsonFile(settingsFilePath, true);
+            builder.AddJsonFile(settingsFilePath, optional: true, reloadOnChange: true);
```
berndbr: **`reloadOnChange: true` installs two never-disposed *recursive* `FileSystemWatcher`s, usually rooted at `AppContext.BaseDirectory`.**

`FileConfigurationSource.ResolveFileProvider()` walks *up* from the settings path until it finds an existing directory, then roots a `PhysicalFileProvider` there. With the default `PluginSystemOptions.PluginSettingsRootPath = "./config"`, that directory typically does not exist (it is absent in this repo's own test output), so the root becomes `AppContext.BaseDirectory` â€” the folder that holds every plugin assembly and plugin subfolder. If `PluginSettingsRootPath` points at an absolute path that does not exist, the walk goes even higher.

`FileConfigurationProvider`'s constructor eagerly calls `ChangeToken.OnChange(() => Source.FileProvider.Watch(Source.Path), â€¦)`, and `PhysicalFilesWatcher` sets `IncludeSubdirectories = true` and `EnableRaisingEvents = true`. So build-time creates **two** recursive watchers (base + `.{Environment}.json`) over the whole deployment tree. The test output directory here already contains 192 directories.

On Linux/containers that is one inotify watch per directory per provider, so a plugin deployment can exhaust `fs.inotify.max_user_watches`; on Windows every file touch anywhere under the deployment wakes the watcher thread. And nothing releases them: `PluginSystemHostContext` is not `IDisposable`, so the `IConfigurationRoot` and its watchers live for the process (and leak per instance in tests).

Consider passing an explicit `PhysicalFileProvider` scoped to the settings directory (creating it if missing), or calling `IConfigurationRoot.Reload()` from the reload trigger instead of watching the tree.

src/PluginSystem/SAF.PluginSystem.Hosting/PluginSystemHostContext.cs
```diff
@@ -46,12 +46,12 @@ private static IConfiguration BuildPluginConfiguration(
...
 
             var filePath = fileSystem.Path.Combine(fileSystem.Path.GetDirectoryName(settingsFilePath)!, fileSystem.Path.GetFileNameWithoutExtension(settingsFilePath));
             var fileExt = fileSystem.Path.GetExtension(settingsFilePath);
             var environmentSettingsFilePath = $"{filePath}.{environment.EnvironmentName}{fileExt}";
-            builder.AddJsonFile(environmentSettingsFilePath, true);
+            builder.AddJsonFile(environmentSettingsFilePath, optional: true, reloadOnChange: true);
```
berndbr: **No `OnLoadException` handler on the now-reloading JSON sources: a transient bad or missing file silently empties `PluginConfiguration`.**

`FileConfigurationProvider.Load(reload: true)` does:

```csharp
catch (Exception e)
{
    if (reload) Data = new Dictionary<string, string?>(...);   // wipes everything
    HandleException(ExceptionDispatchInfo.Capture(e));         // rethrows when OnLoadException is null
}
```

Because `optional: true` only covers the *initial* load, any reload that hits a partially-written file (truncate-then-write from an editor or deploy script) or a momentarily-absent file (delete+rename atomic replace) clears that provider's data and rethrows on the `FileSystemWatcher` callback thread â€” an unhandled exception on a thread-pool thread.

If a `ReloadAsync` lands in that window, every plugin's `ConfigureServices` re-runs against an **empty** configuration: missing connection strings, default endpoints, silently disabled features. Previously a malformed settings file could only fail at startup, where it is a clean, visible failure.

Set `options.OnLoadException` (via `AddJsonFile(Action<JsonConfigurationSource>)`) to log and keep the previous values.

src/PluginSystem/SAF.PluginSystem.Hosting/PluginSystemController.cs
```diff
@@ -0,0 +1,59 @@
...
+        {
+            try
+            {
+                await servicePlugin.StartAsync(cancellationToken).ConfigureAwait(false);
+            }
+            catch (Exception ex)
```
berndbr: **`catch (Exception)` also swallows `OperationCanceledException`, so a cancelled reload logs one bogus per-plugin failure after another.**

When the caller's token is cancelled, `servicePlugin.StartAsync(cancellationToken)` / `StopAsync(cancellationToken)` throw `OperationCanceledException`. Both loops catch it as a generic failure, log `"Failed to start/stop service plug-in during reload."`, and then keep iterating over the remaining plugins with an already-cancelled token â€” so every plugin produces a misleading error entry, and cancellation never actually aborts the operation.

Add `catch (OperationCanceledException) when (cancellationToken.IsCancellationRequested) { throw; }` before the general handler (same applies to `StopServicePluginsAsync` on line 50).

src/PluginSystem/SAF.PluginSystem.Hosting/PluginSystemController.cs
```diff
@@ -0,0 +1,59 @@
...
+            {
+                logger.LogError(ex, "Failed to start service plug-in during reload.");
+            }
+        }
+    }
+
+    private async Task StopServicePluginsAsync(CancellationToken cancellationToken)
+    {
+        foreach (IServicePlugin servicePlugin in GetServicePlugins())
```
berndbr: **The stop pass triggers first-time initialization, so plugins can receive `StopAsync` before they were ever started.**

`GetServicePlugins()` â†’ `pluginServicesContainer.GetPluginServices()` â†’ `InitializeServiceProviders()`. If `ReloadAsync` runs before `ServicePluginHost.StartAsync` has initialized the container (the controller is an independent host singleton, and `ReinitializeAsync_WithoutPriorInitialization_BuildsProviders` shows this path is expected), the stop loop *constructs* every `IServicePlugin` and calls `StopAsync` on instances that were never started â€” a precondition most plugins do not expect.

That entire generation is then immediately thrown away and rebuilt by `ReinitializeAsync`, so plugin construction and any `ConfigureServices` side effects happen twice for nothing.

Skip the stop pass when the container has not been initialized yet (e.g. expose an `IsInitialized` flag, or have `ReinitializeAsync` return whether a previous generation existed).

docs/plugin-system.md
```diff
@@ -401,3 +401,76 @@ sequenceDiagram
...
+
+### What a reload does — and does not — do
+
+- **Rebuilds** every plugin's DI container by re-running `IPluginManifest.ConfigureServices` with the
+  current `PluginConfiguration`, and re-imports the public cross-plugin services.
+- **Disposes** the previously built service providers (and the singletons they own) after the new
+  providers are ready.
+- **Restarts** the `IServicePlugin`s via `ReloadAsync` (stop → reinitialize → start). Per-plugin
+  start/stop failures are logged and do not abort the reload.
+- **Keeps** the assembly load contexts. `PluginAssemblyLoadContext` is **not** collectible, so rebuilding
```
berndbr, line 468: **This "no ALC leak" guarantee rests on an implementation detail the `IPluginAssemblyContainer` contract does not require.**

`ReinitializeAsync` â†’ `BuildProviders()` calls `pluginContainers.SelectMany(s => s.GetPluginManifests())` again on every reload. Reload only avoids leaking assemblies because `PluginAssemblyFolderContainer` happens to cache `_cachedManifests` â€” a private detail documented on that class alone.

`IPluginAssemblyContainer` is public and clearly designed for extension, and its XML doc says nothing about caching or idempotency:

```csharp
/// <summary>Gets the IPluginManifest instances discovered from scanned assemblies.</summary>
IEnumerable<IPluginManifest> GetPluginManifests();
```

A custom container that re-scans on each call will create a fresh `PluginAssemblyLoadContext` per reload â€” and since it is explicitly **not** collectible (as this very bullet states), every reload then leaks a full set of ALCs and assemblies: exactly the leak this design claims to avoid.

Either state the caching requirement on `IPluginAssemblyContainer` (and ideally enforce it, e.g. by having `PluginServicesContainer` cache the manifest list itself on first use), or drop the guarantee from the docs.

src/PluginSystem/SAF.PluginSystem.Hosting/PluginServicesContainer.cs
```diff
@@ -36,6 +36,33 @@ public IServiceProvider GetPublicServices()
         return _publicServicesOnlyCollection.ServiceProvider!;
     }
 
+    public async ValueTask ReinitializeAsync(CancellationToken cancellationToken = default)
+    {
+        List<IServiceProvider> providersToDispose;
+
+        lock (_syncPluginLoading)
```
berndbr, line 50: **The lock is held across arbitrary plugin code, and `System.Threading.Lock` is re-entrant so the `_initialized` guard can be bypassed on the same thread.**

Two problems with holding `_syncPluginLoading` across `BuildProviders()`:

1. **Blocking.** Every concurrent `GetPluginServices()` / `GetPublicServices()` caller â€” including `PluginServiceProvider`, which hosts may call on a request path â€” blocks for the *entire* rebuild: N Ã— `IPluginManifest.ConfigureServices` (arbitrary plugin code, possibly doing I/O) plus N Ã— `BuildServiceProvider()`, plus the `CollectPublicServices` scan which calls `publicServiceTypeRegistry.GetAssemblyNames()` once per `ServiceDescriptor` (a lock acquire + O(m) linear scan each).
2. **Re-entrancy.** `Lock` is re-entrant, so if any manifest's `ConfigureServices` reaches back into the container on the same thread (e.g. resolving `IPluginServiceProvider`), `InitializeServiceProviders()` re-enters the lock, sees `_initialized == true`, returns immediately, and the caller reads the half-built generation â€” `_publicServicesOnlyCollection.ServiceProvider` is still `null` at that point. A different thread doing the same wait can deadlock if the manifest blocks on it.

Build the new generation outside the lock and take it only for the field swap.

src/PluginSystem/SAF.PluginSystem.Hosting/PluginSystemController.cs
```diff
@@ -0,0 +1,59 @@
...
+                await servicePlugin.StopAsync(cancellationToken).ConfigureAwait(false);
+            }
+            catch (Exception ex)
+            {
+                logger.LogError(ex, "Failed to stop service plug-in during reload.");
+            }
+        }
+    }
+
+    private List<IServicePlugin> GetServicePlugins()
```
berndbr: **These three members are a verbatim copy of `ServicePluginHost` â€” that duplication is what caused the missing lifecycle phases.**

`GetServicePlugins()` here is character-for-character identical to `ServicePluginHost.GetServicePlugins()` (line 136-137), and `StartServicePluginsAsync` / `StopServicePluginsAsync` are the same resolve-iterate-try-log-continue loops as `ServicePluginHost.StartAsync` / `StopAsync`, minus the linked `CancellationTokenSource` and minus all four `ILifecycleServicePlugin` phases.

Two owners of the plugin lifecycle will keep drifting â€” it already has. Make `ServicePluginHost` the single owner (extract the start/stop sequence including the lifecycle phases into a shared internal helper, or have the controller drive `ServicePluginHost`) so a reload and a host restart go through exactly the same code path.

src/PluginSystem/SAF.PluginSystem.Hosting/HostApplicationBuilderExtensions.cs
```diff
@@ -43,6 +43,7 @@ public static IPluginSystemHostBuilder AddPluginSystem(this IHostApplicationBuil
         pluginHostBuilder.Services.AddSingleton<IPluginManifestLoader, PluginManifestLoader>();
         pluginHostBuilder.Services.AddSingleton<IPluginServicesContainer, PluginServicesContainer>();
         pluginHostBuilder.Services.AddSingleton<IPluginServiceProvider, PluginServiceProvider>();
+        pluginHostBuilder.Services.AddSingleton<IPluginSystemController, PluginSystemController>();
```
berndbr, line 49: **This new registration is the only one `AddPluginSystem_ShouldAddRequiredServices` does not assert.**

`HostApplicationBuilderExtensionsTests.AddPluginSystem_ShouldAddRequiredServices` is the guard test for this method's registration list, and it covers every other service:

```csharp
Assert.Contains(services, s => s.ServiceType == typeof(IPluginSystemHostContext));
Assert.Contains(services, s => s.ServiceType == typeof(IPluginServicesContainer));
Assert.Contains(services, s => s.ServiceType == typeof(IPluginServiceProvider));
Assert.Contains(services, s => s.ServiceType == typeof(IHostedService) && s.ImplementationType == typeof(ServicePluginHost));
```

`IPluginSystemController` is missing, so deleting this line breaks no test even though `docs/plugin-system.md:422` promises it is "registered automatically by `AddPluginSystem`".

More broadly, none of the reload failure paths are covered: there is no test for a manifest throwing inside `ReinitializeAsync`, for `ReinitializeAsync` throwing inside `ReloadAsync`, for concurrent `ReloadAsync`, or for `ILifecycleServicePlugin` during a reload â€” i.e. none of the defects flagged on this PR would be caught by the new suite.

src/PluginSystem/SAF.PluginSystem.Hosting.Tests/BaseDirectoryFileSystemCollection.cs
```diff
@@ -0,0 +1,14 @@
...
+
+namespace SAF.PluginSystem.Hosting.Tests;
+
+/// <summary>
+/// Groups the tests that enumerate or mutate files directly under <c>AppContext.BaseDirectory</c>
+/// (the plugin-contract assembly scan and the folder-container discovery tests). The collection is
+/// marked non-parallel so one test's recursive directory enumeration cannot race with another test
+/// creating or copying files/directories in the same base directory.
+/// </summary>
+[CollectionDefinition("BaseDirectoryFileSystem", DisableParallelization = true)]
```
berndbr: **Serializing these tests hides the symptom; the cause is that they create directories under `AppContext.BaseDirectory` and never clean up.**

`PluginAssemblyFolderContainerTests` creates `test-plugins-{Guid:N}` directories directly under `AppContext.BaseDirectory` (lines 36, 69, 296) and contains no `Delete`, `finally`, or `IDisposable` cleanup at all â€” while other tests in the same class recursively enumerate `SearchRootPath = AppContext.BaseDirectory` (lines 146, 166, 227, 252, 277), and `PublicServiceTypeRegistryTests` copies an assembly into that same directory.

The residue is already substantial in this working copy:

```
test-plugins* directories under the test output: 105
total directories under the test output:         192
```

That grows by three per test run forever, making the recursive-enumeration tests steadily slower and the shared surface steadily larger â€” `DisableParallelization` only removes the cross-test interleaving, not the growth or the shared mutable state.

The deeper fix is to give each test its own directory outside `AppContext.BaseDirectory` (e.g. under `Path.GetTempPath()`) with cleanup, at which point the collection and its serialization penalty can go away.

Separately: `PluginSystemHostContextTests` and `HostApplicationBuilderExtensionsTests` now build real `PluginSystemHostContext` instances, which â€” via the new `reloadOnChange: true` â€” install recursive `FileSystemWatcher`s over `AppContext.BaseDirectory` that are never disposed, yet neither class is in this collection.

src/PluginSystem/SAF.PluginSystem.Hosting.Contracts/IPluginServicesContainer.cs
```diff
@@ -21,4 +21,21 @@ public interface IPluginServicesContainer
...
+    /// Re-runs each plugin manifest's <see cref="IPluginManifest.ConfigureServices"/> on fresh service
+    /// collections and rebuilds the cross-plugin service wiring, then disposes the previously built
+    /// providers. Because the loaded assemblies / ALCs owned by the plugin containers are left untouched,
+    /// this provides live reconfiguration without a process restart and without leaking assembly load
+    /// contexts. After this call, <see cref="GetPluginServices"/> and <see cref="GetPublicServices"/>
+    /// return the freshly built providers.
+    /// </remarks>
+    /// <param name="cancellationToken">A token that signals cancellation of the reinitialization.</param>
+    /// <returns>A <see cref="ValueTask"/> that completes once the providers have been rebuilt and the previous providers disposed.</returns>
+    ValueTask ReinitializeAsync(CancellationToken cancellationToken = default);
```
berndbr: A destructive lifecycle operation is added to what was a pure read contract, in a shipped package.

`IPluginServicesContainer` had exactly two members, both accessors, and both existing consumers (`PluginServiceProvider`, `ServicePluginHost`) only ever call `GetPluginServices()`. `ReinitializeAsync` rebuilds *and* disposes every plugin provider, so every holder of the read contract now depends on — and can invoke — a method that tears down all other plugins' containers. That is the reader/writer mixing the repo's SOLID rule asks to avoid, and the reload capability already has its own home: `IPluginSystemController`.

Called standalone (which the interface, the XML docs and the docs table all invite), `ReinitializeAsync` disposes the old providers without any `IServicePlugin.StopAsync` and builds new ones without any `StartAsync` — the caller is left with a fully wired container in which every service plugin is dead, and nothing reports it. There is no correct standalone use.

It is also a source and binary break: `SAF.PluginSystem.Hosting.Contracts.csproj` sets no `IsPackable=false`, so this ships. Any external implementer or hand-rolled test double — e.g. the Ext-OPC-UA host on `11.0.0-alpha.8` — fails with `CS0535`, and precompiled implementations get a `TypeLoadException`. The repo has no `PublicAPI.*.txt`, no PublicApiAnalyzers, no `EnablePackageValidation` and no ApiCompat, so nothing flags it, and `docs/migration-10-to-11.md` — the documented breaking-change ledger — has no mention of `ReinitializeAsync`, `ReloadAsync` or `IPluginSystemController` (verified by grep).

Suggestions, in order of preference: keep the mutator off the read contract entirely (internal method, or a narrow internal writer interface on the concrete container), or at minimum give it a default implementation so it stays additive and add a migration-doc entry.

docs/plugin-system.md
```diff
@@ -401,3 +401,76 @@ sequenceDiagram
...
+    participant App as Host / trigger
+    participant PSC as IPluginSystemController
+    participant Cnt as IPluginServicesContainer
+    participant PA  as Service Plugins (old)
+    participant PB  as Service Plugins (new)
+
+    App->>PSC: ReloadAsync
+    PSC->>PA: StopAsync
+    PSC->>Cnt: ReinitializeAsync
+    Note over Cnt: re-run ConfigureServices\nwith fresh PluginConfiguration
```
berndbr: Three small docs issues in the new section, one of which is user-visible on the published site.

1. Line 471 — the `Note over` uses a literal two-character `\n`: `Note over Cnt: re-run ConfigureServices\nwith fresh PluginConfiguration`. Mermaid's sequence renderer breaks note text on real newlines and `<br/>`, not on `\n`, so GitHub Pages renders the backslash-n literally. The pre-existing note at line 392 (`Note over Host: Application running...`) is deliberately single-line, and no other doc file in the repo uses `\n` — use `<br/>` or keep it on one line.

2. Line 458 — `### Reload Sequence Diagram` sits one heading level deeper than its peer `## Lifecycle Sequence Diagram` (line 373), so the two sibling diagrams land at different levels and this one drops out of the h2 page outline.

3. Line 476 — `git diff --check` reports `new blank line at EOF` (`src/.editorconfig`'s `insert_final_newline = false` does not cover `docs/`).

Also worth a thought, not a defect: the reload contract is now stated in the table (419-420), the bullet list (436-445), the diagram (460-475) and again in the XML docs of both new contract members — five copies to keep in lockstep the first time reload behaviour changes. Reducing the table to API/assembly would remove the copy most likely to go stale unnoticed.

src/PluginSystem/SAF.PluginSystem.Hosting.Tests/BaseDirectoryFileSystemCollection.cs
```diff
@@ -0,0 +1,14 @@
+// SPDX-FileCopyrightText: 2017-2026 TRUMPF Laser SE
+//
+// SPDX-License-Identifier: MPL-2.0
+
+namespace SAF.PluginSystem.Hosting.Tests;
+
+/// <summary>
```
berndbr: Six lines of XML rationale on a two-line marker class, and the rationale is already where it belongs — commit ec59c70's message ("Stabilize base-directory filesystem tests: run non-parallel to avoid enumeration race"). This is a test-project marker type, not shipped public API surface, so the Contracts XML-doc convention does not apply here; no other file in the test projects carries a block like this. A single short line (or nothing) would do.

berndbr:
Additional finding on unchanged code that this branch changes the consequences of — `src/PluginSystem/SAF.PluginSystem.Hosting/PublicPluginServiceFactory.cs:100` (not inline-commentable, the file is outside the diff).

**Reload turns a benign shutdown-only leak into a per-reload leak of public cross-plugin singletons.**

`Resolve` caches into `_cachedInstance` (line 63) and, for a type-based descriptor, constructs via `ActivatorUtilities.CreateInstance` (line 100). `ActivatorUtilities.CreateInstance` does not register its result with the provider's disposables list — the owning provider never learns the object exists, and the owner's own patched `sp => CreateInstanceByFactory(sp, sd)` registration is never resolved on this path. The importing side receives a `NonOwningServiceProxy` whose `Dispose` is a no-op.

So a public contract type registered as `AddSingleton<IFoo, Foo>()` and consumed only cross-plugin is disposed by nobody. `ReinitializeAsync` builds a fresh set of factories with fresh empty caches and the previous instance is simply dropped — `DisposeProviderAsync` on the old provider does not reach it.

Concretely: a plugin exposing `AddSingleton<IStorageInfrastructure, SqliteStorage>()` consumed by a sibling plugin leaves one live `SqliteStorage` — connection pool, file handle, background timer — per reload. After k reloads there are k+1 against the same file, and the k-th reload can fail outright with a file-lock `IOException`.

Before this branch the plugin providers were only disposed at process teardown, so the gap was invisible. It is worth deciding whether `PublicPluginServiceFactory` should track and dispose what it creates, or whether the docs' "instances are recreated on every reload" claim needs a carve-out.

Related, same theme: `PluginAssemblyFolderContainer.GetPluginManifests()` caches `_cachedManifests`, so `ConfigureServices` re-runs on the *same* manifest instance — a manifest doing `pluginServices.AddSingleton(_myField)` (as `IHostServiceForwarder.Forward` does) yields the identical instance in the new container, again not recreated and, being an externally-supplied instance, never disposed by MS.DI either.

src/PluginSystem/SAF.PluginSystem.Hosting/PluginSystemController.cs
berndbr: ReloadAsync throws an InvalidOperationException as long as the plugin system has not been initialized/started yet.

src/PluginSystem/SAF.PluginSystem.Hosting/PluginSystemHostContext.cs
berndbr: Applied a PhysicalFileProvider that is scoped to the settings directory.

docs/plugin-system.md
berndbr: Caching plugin manifests in PluginServicesContainer now

src/PluginSystem/SAF.PluginSystem.Hosting.Tests/BaseDirectoryFileSystemCollection.cs
berndbr: File has been removed by fixing the tests

berndbr:
> Additional finding on unchanged code that this branch changes the consequences of — `src/PluginSystem/SAF.PluginSystem.Hosting/PublicPluginServiceFactory.cs:100` (not inline-commentable, the file is outside the diff).
> 
> **Reload turns a benign shutdown-only leak into a per-reload leak of public cross-plugin singletons.**
> 
> `Resolve` caches into `_cachedInstance` (line 63) and, for a type-based descriptor, constructs via `ActivatorUtilities.CreateInstance` (line 100). `ActivatorUtilities.CreateInstance` does not register its result with the provider's disposables list — the owning provider never learns the object exists, and the owner's own patched `sp => CreateInstanceByFactory(sp, sd)` registration is never resolved on this path. The importing side receives a `NonOwningServiceProxy` whose `Dispose` is a no-op.
> 
> So a public contract type registered as `AddSingleton<IFoo, Foo>()` and consumed only cross-plugin is disposed by nobody. `ReinitializeAsync` builds a fresh set of factories with fresh empty caches and the previous instance is simply dropped — `DisposeProviderAsync` on the old provider does not reach it.
> 
> Concretely: a plugin exposing `AddSingleton<IStorageInfrastructure, SqliteStorage>()` consumed by a sibling plugin leaves one live `SqliteStorage` — connection pool, file handle, background timer — per reload. After k reloads there are k+1 against the same file, and the k-th reload can fail outright with a file-lock `IOException`.
> 
> Before this branch the plugin providers were only disposed at process teardown, so the gap was invisible. It is worth deciding whether `PublicPluginServiceFactory` should track and dispose what it creates, or whether the docs' "instances are recreated on every reload" claim needs a carve-out.
> 
> Related, same theme: `PluginAssemblyFolderContainer.GetPluginManifests()` caches `_cachedManifests`, so `ConfigureServices` re-runs on the _same_ manifest instance — a manifest doing `pluginServices.AddSingleton(_myField)` (as `IHostServiceForwarder.Forward` does) yields the identical instance in the new container, again not recreated and, being an externally-supplied instance, never disposed by MS.DI either.

Solved the disposable ownership issue. Caching of the manifest instances is intentionally at the moment. Can be changed later when required.

PR #109, opened 2026-08-06
Allow configruation of external configuration sources

berndbr:
Apply external configuration sources to the plugin settings.
This allows applications to provide settings from different sources and file formats as currently supported in SAF (which is the configured plugin-settings JSON).

### DIFF
--- a/docs/plugin-system.md
+++ b/docs/plugin-system.md
@@ -343,7 +343,7 @@ All plugins share a single plugin settings file, exposed to each manifest as `co
 {PluginSettingsRootPath}/{PluginSettingsFilePath}
 ```
 
-with `PluginSettingsRootPath` defaulting to `.` and `PluginSettingsFilePath` defaulting to `./pluginsettings.json`. An optional environment-specific overlay named `{file}.{EnvironmentName}.json` (e.g. `pluginsettings.Development.json`) is layered on top if present. Both paths come from the `PluginSystem` configuration section.
+with `PluginSettingsRootPath` defaulting to `./config` and `PluginSettingsFilePath` defaulting to `./pluginsettings.json`. An optional environment-specific overlay named `{file}.{EnvironmentName}.json` (e.g. `pluginsettings.Development.json`) is layered on top if present. Both paths come from the `PluginSystem` configuration section.
 
 `PluginSystemOptions` stays a pure data object (paths/patterns only). To add further plugin configuration providers from outside (for example XML, INI, custom sources, or application-specific file names/extensions), use the host builder API:
 
--- a/docs/plugin-system.md
+++ b/docs/plugin-system.md
@@ -357,7 +357,7 @@ var pluginSystemBuilder = builder.AddPluginSystem(options =>
 });
 
 pluginSystemBuilder.AddPluginConfigurationSource(config =>
-    config.AddXmlFile("./config/pluginsettings.myapp", optional: true, reloadOnChange: true));
+    config.AddXmlFile("config/pluginsettings.myapp", optional: true, reloadOnChange: true));
 
 pluginSystemBuilder.AddPluginConfigurationSource(config =>
     config.AddXmlFile(
--- a/docs/plugin-system.md
+++ b/docs/plugin-system.md
@@ -350,37 +350,48 @@ with `PluginSettingsRootPath` defaulting to `./config` and `PluginSettingsFilePa
 ```csharp
 var builder = Host.CreateApplicationBuilder(args);
 
-var pluginConfigRootPath = Path.Combine(AppContext.BaseDirectory, "config");
-
 var pluginSystemBuilder = builder.AddPluginSystem(options =>
 {
-    options.PluginSettingsRootPath = pluginConfigRootPath;
     options.PluginSettingsFilePath = "./pluginsettings.json";
 });
 
-pluginSystemBuilder.AddPluginConfigurationSource(config =>
-    config.AddXmlFile(Path.Combine(pluginConfigRootPath, "pluginsettings.myapp"), optional: true, reloadOnChange: true));
+pluginSystemBuilder.AddPluginConfigurationSource(source =>
+{
+    var extension = ".myapp";
+    var overlayFileName = $"{Path.GetFileNameWithoutExtension(source.SettingsFileName)}.{source.EnvironmentName}" +
+        Path.GetExtension(source.SettingsFileName);
 
-pluginSystemBuilder.AddPluginConfigurationSource(config =>
-    config.AddXmlFile(
-        Path.Combine(pluginConfigRootPath, $"pluginsettings.{builder.Environment.EnvironmentName}.myapp"),
-        optional: true,
-        reloadOnChange: true));
+    source.Builder.AddXmlFile(xml =>
+    {
+        xml.FileProvider = source.SettingsFileProvider;
+        xml.Path = Path.ChangeExtension(source.SettingsFileName, extension);
+        xml.Optional = true;
+        xml.ReloadOnChange = true;
+        xml.OnLoadException = source.OnLoadException;
+    });
+
+    source.Builder.AddXmlFile(xml =>
+    {
+        xml.FileProvider = source.SettingsFileProvider;
+        xml.Path = Path.ChangeExtension(overlayFileName, extension);
+        xml.Optional = true;
+        xml.ReloadOnChange = true;
+        xml.OnLoadException = source.OnLoadException;
+    });
+});
 ```
 
-Additional providers are appended after the default plugin JSON sources. Configuration precedence follows normal .NET rules (later providers override earlier ones).
+The callback receives a `PluginConfigurationSourceContext` with everything the built-in plugin settings
+pipeline already resolved: `SettingsFileProvider` (the `IFileProvider` scoped to the resolved settings
+directory — the same instance the default plugin JSON files use), `SettingsFileName` (e.g.
+`pluginsettings.json`), `EnvironmentName`, and `OnLoadException` (the shared handler that ignores a failed
+load and logs a warning instead of crashing host startup or silently wiping values on reload). Building
+sources through `source.SettingsFileProvider` keeps them rooted at the same directory as the default plugin
+JSON regardless of how `PluginSettingsRootPath` resolves — there is no separate path to keep in sync.
 
-> **`PluginSettingsRootPath` and custom source paths resolve independently.** The default plugin JSON is
-> located by resolving `PluginSettingsRootPath` (after `Environment.ExpandEnvironmentVariables`) against
-> `AppContext.BaseDirectory` — and `PluginSettingsRootPath` may itself be absolute (e.g.
-> `%ProgramData%/MyApp/config`). Custom sources added via `AddPluginConfigurationSource`, in contrast,
-> resolve through the configuration builder's file provider, which defaults to a
-> `PhysicalFileProvider(AppContext.BaseDirectory)` — it has no knowledge of `PluginSettingsRootPath` at all.
-> A relative path like `"./config"` used for both happens to resolve to the same place, but the two only
-> agree by coincidence — if `PluginSettingsRootPath` is changed to an absolute path elsewhere, the plugin
-> JSON and the custom sources silently start reading from different directories (worse yet, with
-> `optional: true` a miss stays silent). Building one absolute root and reusing it for both, as above, keeps
-> them in sync regardless of what `PluginSettingsRootPath` resolves to.
+The callback runs exactly once, during `IPluginSystemHostContext` construction; any exception it throws
+propagates into host startup. Additional providers are appended after the default plugin JSON sources.
+Configuration precedence follows normal .NET rules (later providers override earlier ones).
 
 Because the file is shared, plugins keep their settings under distinct top-level sections (the built-in messaging/storage plugins use `Messaging`, `Redis`, `Nats`, `LiteDb`, `SQLite`, `MessageRouting`).
 
--- a/docs/plugin-system.md
+++ b/docs/plugin-system.md
@@ -345,6 +345,29 @@ All plugins share a single plugin settings file, exposed to each manifest as `co
 
 with `PluginSettingsRootPath` defaulting to `.` and `PluginSettingsFilePath` defaulting to `./pluginsettings.json`. An optional environment-specific overlay named `{file}.{EnvironmentName}.json` (e.g. `pluginsettings.Development.json`) is layered on top if present. Both paths come from the `PluginSystem` configuration section.
 
+`PluginSystemOptions` stays a pure data object (paths/patterns only). To add further plugin configuration providers from outside (for example XML, INI, custom sources, or application-specific file names/extensions), use the host builder API:
+
+```csharp
+var builder = Host.CreateApplicationBuilder(args);
+
+var pluginSystemBuilder = builder.AddPluginSystem(options =>
+{
+    options.PluginSettingsRootPath = "./config";
+    options.PluginSettingsFilePath = "./pluginsettings.json";
+});
+
+pluginSystemBuilder.AddPluginConfigurationSource(config =>
+    config.AddXmlFile("./config/pluginsettings.myapp", optional: true, reloadOnChange: true));
+
+pluginSystemBuilder.AddPluginConfigurationSource(config =>
+    config.AddXmlFile(
+        $"./config/pluginsettings.{builder.Environment.EnvironmentName}.myapp",
+        optional: true,
+        reloadOnChange: true));
+```
+
+Additional providers are appended after the default plugin JSON sources. Configuration precedence follows normal .NET rules (later providers override earlier ones).
+
 Because the file is shared, plugins keep their settings under distinct top-level sections (the built-in messaging/storage plugins use `Messaging`, `Redis`, `Nats`, `LiteDb`, `SQLite`, `MessageRouting`).
 
 Access your section via `context.PluginConfiguration` in `ConfigureServices`:
@@ -369,6 +392,8 @@ Example `pluginsettings.json`:
 ```
 
 > A plugin may also fall back to host configuration (`context.HostConfiguration`) — the built-in plugins read their section from `PluginConfiguration` first and fall back to `HostConfiguration`. In the simplest setups you can point `PluginSettingsFilePath` at the host's `appsettings.json` and keep everything in one file.
+>
+> If you use `AddXmlFile(...)`, add the package `Microsoft.Extensions.Configuration.Xml` to the host project.
 
 ### Change tracking
 
--- a/docs/plugin-system.md
+++ b/docs/plugin-system.md
@@ -350,23 +350,37 @@ with `PluginSettingsRootPath` defaulting to `.` and `PluginSettingsFilePath` def
 ```csharp
 var builder = Host.CreateApplicationBuilder(args);
 
+var pluginConfigRootPath = Path.Combine(AppContext.BaseDirectory, "config");
+
 var pluginSystemBuilder = builder.AddPluginSystem(options =>
 {
-    options.PluginSettingsRootPath = "./config";
+    options.PluginSettingsRootPath = pluginConfigRootPath;
     options.PluginSettingsFilePath = "./pluginsettings.json";
 });
 
 pluginSystemBuilder.AddPluginConfigurationSource(config =>
-    config.AddXmlFile("config/pluginsettings.myapp", optional: true, reloadOnChange: true));
+    config.AddXmlFile(Path.Combine(pluginConfigRootPath, "pluginsettings.myapp"), optional: true, reloadOnChange: true));
 
 pluginSystemBuilder.AddPluginConfigurationSource(config =>
     config.AddXmlFile(
-        $"./config/pluginsettings.{builder.Environment.EnvironmentName}.myapp",
+        Path.Combine(pluginConfigRootPath, $"pluginsettings.{builder.Environment.EnvironmentName}.myapp"),
         optional: true,
         reloadOnChange: true));
 ```
 
 Additional providers are appended after the default plugin JSON sources. Configuration precedence follows normal .NET rules (later providers override earlier ones).
+
+> **`PluginSettingsRootPath` and custom source paths resolve independently.** The default plugin JSON is
+> located by resolving `PluginSettingsRootPath` (after `Environment.ExpandEnvironmentVariables`) against
+> `AppContext.BaseDirectory` — and `PluginSettingsRootPath` may itself be absolute (e.g.
+> `%ProgramData%/MyApp/config`). Custom sources added via `AddPluginConfigurationSource`, in contrast,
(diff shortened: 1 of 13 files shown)


docs/plugin-system.md
```diff
@@ -345,6 +345,29 @@ All plugins share a single plugin settings file, exposed to each manifest as `co
...
+var builder = Host.CreateApplicationBuilder(args);
+
+var pluginSystemBuilder = builder.AddPluginSystem(options =>
+{
+    options.PluginSettingsRootPath = "./config";
+    options.PluginSettingsFilePath = "./pluginsettings.json";
+});
+
+pluginSystemBuilder.AddPluginConfigurationSource(config =>
+    config.AddXmlFile("./config/pluginsettings.myapp", optional: true, reloadOnChange: true));
```
berndbr: **[correctness] The `./` prefix silently disables `reloadOnChange` — the watcher never fires.**

`PhysicalFileProvider.Watch()` only trims leading *separators*, so the change token is registered under the key `./config/pluginsettings.myapp`, while the file-watcher event reports the relative path as `config/pluginsettings.myapp`. The keys never match, so the token is never cancelled and the provider is never reloaded — silently, with no error.

Verified on net10.0 / `Microsoft.Extensions.Configuration 10.0.0`:

```
[config] path='./config/pluginsettings.json' initial='before'
[config] path='./config/pluginsettings.json' afterEdit='before'   <-- never reloaded
[config] path='config/pluginsettings.json'   initial='before'
[config] path='config/pluginsettings.json'   afterEdit='after'    <-- reloaded
```

This is exactly the invariant `PluginSystemHostContext` already protects for the built-in JSON sources (it passes a bare file name plus a directory-scoped `PhysicalFileProvider`), and which `PluginSystemHostContextTests.BuildPluginConfiguration_EnablesReloadOnChange_ForPluginSettingsFiles` asserts:

```csharp
Assert.All(fileProviders, provider => Assert.DoesNotContain(fileSystem.Path.DirectorySeparatorChar, provider.Source.Path!));
```

Same problem on line 364 of this file and on `docs/saf-host.md:135` / `:139`.

```suggestion
    config.AddXmlFile("config/pluginsettings.myapp", optional: true, reloadOnChange: true));
```

src/PluginSystem/SAF.PluginSystem.Hosting/PluginSystemHostContext.cs
```diff
@@ -82,10 +100,23 @@ private static (IConfigurationRoot ConfigurationRoot, PhysicalFileProvider? Sett
...
+    }
+    
+    private static void AddCustomPluginConfigurationSources(
+        IConfigurationBuilder builder,
+        IEnumerable<Action<IConfigurationBuilder>> configurePluginConfigurationSources)
+    {
+        ArgumentNullException.ThrowIfNull(builder);
+        ArgumentNullException.ThrowIfNull(configurePluginConfigurationSources);
+
+        foreach (var configureSource in configurePluginConfigurationSources)
```
berndbr, line 146: **[correctness] Custom sources bypass the `OnLoadException` guard — a malformed file now crashes host startup, and on reload it silently wipes the source's values.**

Every default source goes through `AddJsonFile(...)`, which sets `Optional = true` **and** an `OnLoadException` handler that sets `context.Ignore = true` and logs a warning. Sources added through `AddPluginConfigurationSource` get neither, so the failure modes documented in `docs/plugin-system.md:405` ("A malformed or half-written settings file does not fail the host … the previously loaded values stay in effect") do not hold for them.

Verified on net10.0:

```
BUILD-TIME THROWS: InvalidDataException / inner=FormatException   // even with optional: true
RELOAD: initial=ok
RELOAD: survived, value now=                                       // values silently dropped, nothing logged
```

* **Startup**: a malformed `pluginsettings.myapp` makes `builder.Build()` throw `InvalidDataException` inside the `AddSingleton<IPluginSystemHostContext>` factory → `ServicePluginHost` fails to construct → the host dies at `StartAsync`. `optional: true` does not help; it only covers a *missing* file.
* **Runtime reload**: `FileConfigurationProvider.Load(reload: true)` clears `Data` before parsing and re-clears it on failure; the rethrow happens on a `Task.Factory.StartNew` continuation, so it is swallowed as an unobserved task exception. Result: every key from that source vanishes from `context.PluginConfiguration` with **no log entry at all**, and plugins bound via `IOptionsMonitor<T>` silently fall back to defaults.

Consider applying the same treatment the defaults get — e.g. wrap the builder handed to the callback so file sources inherit the `OnLoadException` handler, or at minimum post-process `builder.Sources` for `FileConfigurationSource`s with a null `OnLoadException` and attach the shared handler before `Build()`.

src/PluginSystem/SAF.PluginSystem.Hosting/PluginSystemHostContext.cs
```diff
@@ -43,18 +45,34 @@ public void Dispose()
...
-        IFileSystem fileSystem)
+        IFileSystem fileSystem,
+        IEnumerable<Action<IConfigurationBuilder>> configurePluginConfigurationSources)
     {
         var builder = new ConfigurationBuilder();
 
+        var settingsFileProvider = AddDefaultPluginConfigurationSources(builder, logger, options, environment, fileSystem);
+        AddCustomPluginConfigurationSources(builder, configurePluginConfigurationSources);
+
+        return (builder.Build(), settingsFileProvider);
```
berndbr: **[correctness] The settings `PhysicalFileProvider` (and its `FileSystemWatcher`) leaks whenever a custom source or `Build()` throws.**

`AddDefaultPluginConfigurationSources` constructs `new PhysicalFileProvider(settingsDirectoryPath)` (line 89) and returns it. User code then runs at line 58, and `builder.Build()` at line 60. If either throws — which is now a realistic path, see the `OnLoadException` comment above — the exception escapes the `PluginSystemHostContext` constructor, `_pluginSettingsFileProvider` is never assigned, and `Dispose()` can never run. The provider, its `PhysicalFilesWatcher` and the underlying `FileSystemWatcher` handle stay alive until process exit.

Before this PR no user-supplied code ran between provider creation and `Build()`, and the JSON sources were unable to throw (optional + ignored load exceptions), so the window did not exist.

A `try`/`catch` around lines 58-60 that disposes `settingsFileProvider` and rethrows would close it.

src/PluginSystem/SAF.PluginSystem.Hosting/PluginSystemHostContext.cs
```diff
@@ -82,10 +100,23 @@ private static (IConfigurationRoot ConfigurationRoot, PhysicalFileProvider? Sett
         var environmentSettingsFileName = $"{fileSystem.Path.GetFileNameWithoutExtension(settingsFileName)}.{environment.EnvironmentName}{fileSystem.Path.GetExtension(settingsFileName)}";
         AddJsonFile(builder, logger, settingsFileProvider, environmentSettingsFileName);
 
-        return (builder.Build(), settingsFileProvider);
+        return settingsFileProvider;
+    }
+    
+    private static void AddCustomPluginConfigurationSources(
```
berndbr, line 138: **[correctness] `Dispose()` does not release file providers created by custom sources — each one holds a *recursive* `FileSystemWatcher` over `AppContext.BaseDirectory`.**

`Dispose()` only disposes `_pluginConfigurationRoot` and `_pluginSettingsFileProvider`. `ConfigurationRoot.Dispose()` disposes the *providers*, but `FileConfigurationProvider.Dispose()` only releases its change-token registration — it never touches `Source.FileProvider`.

For the documented usage (`config.AddXmlFile("...", reloadOnChange: true)` with no explicit provider), `FileConfigurationSource.EnsureDefaults` falls back to `builder.GetFileProvider()`, which returns a **new** `PhysicalFileProvider(AppContext.BaseDirectory)` per source. Verified on net10.0:

```
provider[1] root='...\bin\Debug\net10.0\'
provider[2] root='...\bin\Debug\net10.0\'
custom src1 provider == src2 provider: False
custom PhysicalFileProvider disposed after ConfigurationRoot.Dispose(): False
```

So the two sources in the docs give you two undisposed `PhysicalFileProvider`s, each with an `IncludeSubdirectories = true` watcher on the whole application base directory — firing on every plugin DLL, log file and temp file written there. In integration tests or any scenario that builds and disposes hosts repeatedly this is a handle/thread leak.

Either track the providers the custom sources end up using and dispose them, or seed `builder.SetFileProvider(settingsFileProvider)` so custom file sources reuse the provider this class already owns and disposes.

docs/plugin-system.md
```diff
@@ -345,6 +345,29 @@ All plugins share a single plugin settings file, exposed to each manifest as `co
...
 with `PluginSettingsRootPath` defaulting to `.` and `PluginSettingsFilePath` defaulting to `./pluginsettings.json`. An optional environment-specific overlay named `{file}.{EnvironmentName}.json` (e.g. `pluginsettings.Development.json`) is layered on top if present. Both paths come from the `PluginSystem` configuration section.
 
+`PluginSystemOptions` stays a pure data object (paths/patterns only). To add further plugin configuration providers from outside (for example XML, INI, custom sources, or application-specific file names/extensions), use the host builder API:
+
+```csharp
+var builder = Host.CreateApplicationBuilder(args);
+
+var pluginSystemBuilder = builder.AddPluginSystem(options =>
+{
+    options.PluginSettingsRootPath = "./config";
```
berndbr: **[docs] The example hardcodes `./config/` while `PluginSettingsRootPath` is the configured root — the two silently diverge.**

The sample sets `options.PluginSettingsRootPath = "./config"` and then repeats `./config/` in both `AddXmlFile` calls. Those are resolved by two completely different mechanisms:

* the default JSON file → `CreateAbsolutePath(AppContext.BaseDirectory, Path.Combine(environment.PluginSettingsRootPath, options.PluginSettingsFilePath))`, where `PluginSettingsRootPath` has already been through `Environment.ExpandEnvironmentVariables(...)` and may be **absolute**;
* the custom XML files → `builder.GetFileProvider()` → `new PhysicalFileProvider(AppContext.BaseDirectory)`, i.e. always relative to the base directory.

They only agree here by coincidence. As soon as someone sets `"PluginSettingsRootPath": "%ProgramData%/MyApp/config"` or an absolute path (both supported, see `HostApplicationBuilderExtensions.cs:29`), the plugin JSON is read from one place and the custom XML from another — and because the docs recommend `optional: true`, the miss is completely silent.

Worth either using an absolute path built from the same root in the sample, or noting explicitly that custom source paths resolve against `AppContext.BaseDirectory` and not `PluginSettingsRootPath`.

docs/saf-host.md
```diff
@@ -131,6 +131,15 @@ builder.AddSafHost(pluginSystemOptions =>
 })
 .ConfigurePluginSystem(ps =>
 {
+    ps.AddPluginConfigurationSource(config =>
+        config.AddXmlFile("./config/pluginsettings.myapp", optional: true, reloadOnChange: true));
```
berndbr: **[correctness] Same `./`-prefix problem as `docs/plugin-system.md:360`: `reloadOnChange: true` never fires here.**

`PhysicalFileProvider.Watch()` trims leading separators but not `./`, so the change token is keyed `./config/pluginsettings.myapp` while watcher events report `config/pluginsettings.myapp`. The file is read once at startup and never reloaded — silently. Applies to line 139 as well.

```suggestion
        config.AddXmlFile("config/pluginsettings.myapp", optional: true, reloadOnChange: true));
```

src/PluginSystem/SAF.PluginSystem.Hosting.Tests/PluginSystemHostContextTests.cs
```diff
@@ -153,4 +153,58 @@ public void BuildPluginConfiguration_EnablesReloadOnChange_ForPluginSettingsFile
...
+            builder.AddInMemoryCollection(new Dictionary<string, string?>
+            {
+                ["Key"] = "OverriddenByCustomSource",
+            }),
+        };
+        // PluginSystemHostContext builds its configuration from settings files on disk, so a real file system is used.
+        var fileSystem = new RealFileSystem();
+
+        // Act
+        var context = new PluginSystemHostContext(logger, environment, hostConfiguration, options, fileSystem, configureSources);
```
berndbr: **[test-hygiene] Missing `using` — this test leaks a real `PhysicalFileProvider` + `FileSystemWatcher`.**

All six pre-existing tests in this file use `using var context = new PluginSystemHostContext(...)`. Both new tests (line 177 and line 205) drop it. This one is the expensive case: with `PluginSettingsFilePath = "settings.json"` and `PluginSettingsRootPath = "./test-plugin-configs"` it builds a real `PhysicalFileProvider` with `ReloadOnChange = true` over the output directory, and `PluginSystemHostContext.Dispose()` is never called — the watcher thread and directory handle survive for the whole test run.

```suggestion
        using var context = new PluginSystemHostContext(logger, environment, hostConfiguration, options, fileSystem, configureSources);
```

src/PluginSystem/SAF.PluginSystem.Hosting.Tests/HostApplicationBuilderExtensionsTests.cs
```diff
@@ -101,4 +101,37 @@ public void AddPluginSystem_ConfiguresPluginSystemHostEnvironment()
...
+        {
+            options.PluginSettingsFilePath = string.Empty;
+        });
+        pluginSystemHostBuilder.AddPluginConfigurationSource(configurationBuilder =>
+            configurationBuilder.AddInMemoryCollection(new Dictionary<string, string?>
+            {
+                ["Custom:Setting"] = "ValueFromOutside",
+            }));
+
+        var serviceProvider = services.BuildServiceProvider();
```
berndbr: **[test-hygiene] Missing `using` on the service provider — the singleton `PluginSystemHostContext` is never disposed.**

The two neighbouring tests in this file (lines 73 and 95) both use `using var serviceProvider = services.BuildServiceProvider();`. Dropping it here means the container — and the `IDisposable` `PluginSystemHostContext` singleton it created — is left to the GC.

```suggestion
        using var serviceProvider = services.BuildServiceProvider();
```

src/PluginSystem/SAF.PluginSystem.Hosting/PluginSystemHostBuilderExtensions.cs
```diff
@@ -5,13 +5,33 @@
...
+    /// </summary>
+    /// <param name="hostBuilder">The plugin system host builder.</param>
+    /// <param name="configureSource">The callback that appends one or more providers to the plugin configuration builder.</param>
+    /// <returns>The same <see cref="IPluginSystemHostBuilder"/> instance for chaining.</returns>
+    public static IPluginSystemHostBuilder AddPluginConfigurationSource(
+        this IPluginSystemHostBuilder hostBuilder,
+        Action<IConfigurationBuilder> configureSource)
+    {
+        ArgumentNullException.ThrowIfNull(hostBuilder);
+        ArgumentNullException.ThrowIfNull(configureSource);
```
berndbr, line 34: **[test-coverage] The new guard clauses and the multi-source ordering guarantee are untested.**

`PluginSystemHostBuilderExtensionsTests.cs` — the dedicated test file for this class — was not touched by this PR. The sibling method has explicit tests for exactly these branches:

```csharp
[Fact] public void AddPluginAssemblyFolderContainer_ThrowsIfHostBuilderEqualsNull()
[Fact] public void AddPluginAssemblyFolderContainer_WithConfiguration_ThrowsIfHostBuilderEqualsNull()
```

Missing for `AddPluginConfigurationSource`:
* `hostBuilder` null → `ArgumentNullException`
* `configureSource` null → `ArgumentNullException`
* **two or more** `AddPluginConfigurationSource` calls preserve registration order — this is the guarantee the docs make ("later providers override earlier ones") and the only test that touches ordering (`BuildPluginConfiguration_ShouldApplyCustomSourcesAfterDefaultSources`) bypasses the public API by calling the constructor directly with a hand-built list.

docs/plugin-system.md
```diff
@@ -345,6 +345,29 @@ All plugins share a single plugin settings file, exposed to each manifest as `co
 
 with `PluginSettingsRootPath` defaulting to `.` and `PluginSettingsFilePath` defaulting to `./pluginsettings.json`. An optional environment-specific overlay named `{file}.{EnvironmentName}.json` (e.g. `pluginsettings.Development.json`) is layered on top if present. Both paths come from the `PluginSystem` configuration section.
 
+`PluginSystemOptions` stays a pure data object (paths/patterns only). To add further plugin configuration providers from outside (for example XML, INI, custom sources, or application-specific file names/extensions), use the host builder API:
```
berndbr, line 348: **[docs] The paragraph directly above this addition states the wrong default, and the new sample contradicts it.**

Line 346 says *"with `PluginSettingsRootPath` defaulting to `.`"*, but `PluginSystemOptions.PluginSettingsRootPath` is `"./config"` (`src/PluginSystem/SAF.PluginSystem.Hosting/PluginSystemOptions.cs:21`), and `docs/saf-host.md:70` documents the default as `./config`. The new sample right below sets `"./config"` too, so the doc now disagrees with itself two lines apart.

`docs/migration-10-to-11.md:273` repeats the same wrong default.

src/PluginSystem/SAF.PluginSystem.Hosting/HostApplicationBuilderExtensions.cs
```diff
@@ -36,7 +36,9 @@ public static IPluginSystemHostBuilder AddPluginSystem(this IHostApplicationBuil
             var environment = sp.GetRequiredService<IPluginSystemHostEnvironment>();
             var logger = sp.GetRequiredService<ILogger<PluginSystemHostContext>>();
             var fileSystem = sp.GetRequiredService<IFileSystem>();
-            return new PluginSystemHostContext(logger, environment, hostAppBuilder.Configuration, options.Value, fileSystem);
+            var configurationSourcesOptions = sp.GetService<IOptions<PluginConfigurationSourcesOptions>>();
```
berndbr: **[simplification] Three layers of null handling for a value that cannot be null.**

`pluginHostBuilder.Services.Configure(configure)` on line 22 calls `AddOptions()`, which registers `IOptions<>` as an **open generic** singleton. `GetService<IOptions<PluginConfigurationSourcesOptions>>()` therefore always resolves (returning an options instance with an empty `ConfigureSources` list when nobody called `AddPluginConfigurationSource`). The `GetService` + `?.` here and the `?? []` in the constructor are all dead.

```suggestion
            var configureSources = sp.GetRequiredService<IOptions<PluginConfigurationSourcesOptions>>().Value.ConfigureSources;
            return new PluginSystemHostContext(logger, environment, hostAppBuilder.Configuration, options.Value, fileSystem, configureSources);
```

src/PluginSystem/SAF.PluginSystem.Hosting/PluginSystemHostContext.cs
```diff
@@ -82,10 +100,23 @@ private static (IConfigurationRoot ConfigurationRoot, PhysicalFileProvider? Sett
...
+        IConfigurationBuilder builder,
+        IEnumerable<Action<IConfigurationBuilder>> configurePluginConfigurationSources)
+    {
+        ArgumentNullException.ThrowIfNull(builder);
```
berndbr: **[simplification] Dead guard clauses in `private static` helpers.**

`AddCustomPluginConfigurationSources` has exactly one caller, `BuildPluginConfiguration`, which passes a freshly constructed `ConfigurationBuilder` and a sequence already coalesced with `?? []`. Neither argument can be null, so both `ThrowIfNull` calls are unreachable — as is the `ThrowIfNull(builder)` at line 70. They also count against the coverage target because no test can ever hit them.

```suggestion
    private static void AddCustomPluginConfigurationSources(
        IConfigurationBuilder builder,
        IEnumerable<Action<IConfigurationBuilder>> configurePluginConfigurationSources)
    {
        foreach (var configureSource in configurePluginConfigurationSources)
```

src/PluginSystem/SAF.PluginSystem.Hosting/PluginSystemHostContext.cs
```diff
@@ -21,12 +21,14 @@ public PluginSystemHostContext(
         IPluginSystemHostEnvironment environment,
         IConfigurationManager hostConfiguration,
         PluginSystemOptions options,
-        IFileSystem fileSystem)
+        IFileSystem fileSystem,
+        IEnumerable<Action<IConfigurationBuilder>>? configurePluginConfigurationSources = null)
```
berndbr: **[api-compat] Adding an optional parameter to this public constructor is a binary-breaking change.**

`PluginSystemHostContext` is `public sealed` in a packable assembly (`Directory.Build.props` supplies full package metadata and the project does not set `IsPackable=false`). Optional arguments are baked in at the *call site*, so any assembly already compiled against the 5-argument constructor resolves a member token that no longer exists and fails with `MissingMethodException` at runtime — a recompile is required.

If binary compatibility matters for the 11.x line, keeping the old 5-argument constructor as an overload that forwards with `null` costs two lines and removes the break entirely.

src/PluginSystem/SAF.PluginSystem.Hosting/PluginSystemHostBuilderExtensions.cs
```diff
@@ -5,13 +5,33 @@
...
+using Microsoft.Extensions.Configuration;
 using Microsoft.Extensions.DependencyInjection;
 using Microsoft.Extensions.Logging;
 using Microsoft.Extensions.Options;
 using System.IO.Abstractions;
 
 public static class PluginSystemHostBuilderExtensions
 {
+    /// <summary>
+    /// Appends a custom plugin configuration source to the plugin configuration builder pipeline.
```
berndbr, line 16: **[altitude] The extension point is one level too shallow for its own documented use case.**

The motivating scenario in `docs/plugin-system.md:348` is *"application-specific file names/extensions"* for the same plugin settings file. But `AddDefaultPluginConfigurationSources` already owns everything that scenario needs — the resolved settings directory, a `PhysicalFileProvider` scoped to it, the `{name}.{EnvironmentName}{ext}` overlay convention, and the shared `OnLoadException` handler — and none of it is reachable from an `Action<IConfigurationBuilder>`.

The docs show the consequence: two nearly identical calls per format, each re-deriving the environment-overlay name by hand and re-hardcoding `./config/`, and neither inheriting the error handling or the reload semantics the built-in pipeline guarantees (see the separate comments on those three points).

A hook that hands the callback a small context — settings `IFileProvider`, environment name, base settings file name, and the shared load-exception handler — would let a caller express "same file, different extension" in one call and would keep all four defects out of user code. Also worth documenting on the XML summary: the callback runs exactly once, during `IPluginSystemHostContext` construction, and any exception it throws propagates into host startup.

src/PluginSystem/SAF.PluginSystem.Hosting/PluginSystemHostContext.cs
berndbr: Binary compatibility doesn't matter for major version 11.0

Issue #15, opened 2020-10-28
The ScopeId set in appsettings.json is not applied to CDE


berndbr:
While using the default CDE modules without any specific crypto library a ScopeId set in the appsettings.json will not be applied to the CDE because of changed behaviour of OSS CDE version >= 5.x.

With that error the CDE always runs without unscoped.

PR #110, opened 2026-08-10
Verify plug-in assemblies before loading them

berndbr:
Introduce possibilty to verify plug-in assemblies before they get loaded.
Add implementations that validates strong name keys or code signing certificates.


src/PluginSystem/SAF.PluginSystem.Hosting.Extensions/Authenticode/AuthenticodeSignatureReader.cs
```diff
@@ -0,0 +1,242 @@
...
+            ? signedData
+            : null;
+    }
+
+    private static FileStream OpenTrustSnapshot(ReadOnlyMemory<byte> assemblyBytes, string path)
+    {
+        var options = new FileStreamOptions
+        {
+            Mode = FileMode.CreateNew,
+            Access = FileAccess.ReadWrite,
```
berndbr: **Critical — `RequireValidDigitalSignature = true` rejects *every* plugin on Windows, including correctly signed and fully trusted ones.**

The trust snapshot is kept open with `FileAccess.ReadWrite` while `WindowsAuthenticodeTrustVerifier.IsTrusted(trustSnapshot.Name, ...)` runs. `WinVerifyTrust` opens the file itself, and Windows refuses that open while a *write-capable* handle exists on it, so `WinVerifyTrust` never returns `S_OK`, `isTrusted` is always `false`, and `ReadSignatureCore` returns `HasValidDigitalSignature = isTrusted && signatureCoversFile` = `false`.

This is the production path: `PluginAssemblyFolderContainer` always supplies `AssemblyBytes`, so `DigitalSignaturePluginAssemblyValidator` always picks the `ReadSignature(ReadOnlyMemory<byte>)` overload.

Verified locally against a trusted Authenticode-signed binary (`C:\Windows\System32\kernel32.dll`) with `RequireValidDigitalSignature = true`:

| validation context | result |
|---|---|
| `new PluginAssemblyValidationContext(path, name)` (path overload) | accepted |
| `new PluginAssemblyValidationContext(path, name, bytes)` (what the container builds) | **rejected** - "assembly does not have a valid digital signature" |

Isolated to the held handle - same file, verified by path while a handle is held open:

| held handle | outcome |
|---|---|
| `FileAccess.ReadWrite` + `FileShare.Read` (current code) | rejected |
| `FileAccess.ReadWrite` + `FileShare.ReadWrite` | rejected |
| `FileAccess.Read` + `FileShare.Read` | **accepted** |
| `FileAccess.Read` + `FileShare.Read or FileShare.Delete` | **accepted** |

An `AllowedSignerThumbprints`-only configuration still works, because `signatureCoversFile` falls back to the PE-hash check - only the `RequireValidDigitalSignature` switch is dead.

Fix: write and flush the snapshot through a write handle, close it, then reopen read-only (`FileAccess.Read`, `FileShare.Read | FileShare.Delete`) and hold that handle across the verification to keep the anti-TOCTOU guarantee.

src/PluginSystem/SAF.PluginSystem.Hosting.Extensions/Authenticode/CrossPlatformAuthenticodeTrustVerifier.cs
```diff
@@ -0,0 +1,38 @@
...
+
+    public bool IsTrusted(string assemblyPath, X509Certificate2 signerCertificate)
+    {
+        using var chain = new X509Chain();
+        chain.ChainPolicy.RevocationMode = X509RevocationMode.NoCheck;
+        chain.ChainPolicy.RevocationFlag = X509RevocationFlag.ExcludeRoot;
+        chain.ChainPolicy.VerificationFlags = X509VerificationFlags.NoFlag;
+        chain.ChainPolicy.ApplicationPolicy.Add(new Oid(CodeSigningEnhancedKeyUsageOid));
+
+        return chain.Build(signerCertificate);
```
berndbr, line 75: **The CMS-embedded intermediate certificates are never given to the chain, so `Build` returns `PartialChain` and legitimately signed plugins are rejected on Linux/macOS and on any host that cannot fetch AIA.**

Authenticode signatures carry the leaf and the intermediate CA certificates inside the PKCS#7 blob; OS trust stores normally hold only roots. Without `chain.ChainPolicy.ExtraStore.AddRange(cms.Certificates)` the chain cannot be completed. On Linux/macOS .NET has no Windows CryptoAPI intermediate cache, and air-gapped or egress-restricted industrial hosts cannot download via AIA either.

Verified with the leaf certificate of `kernel32.dll` and `DisableCertificateDownloads = true` (simulating no AIA):

| policy | Build() | elements | status |
|---|---|---|---|
| as written (no ExtraStore) | `False` | 1 | `PartialChain, NotTimeValid` |
| plus `ExtraStore = cms.Certificates` | `False` | 3 | `NotTimeValid` |

Note the interface makes this unfixable as designed: `IAuthenticodeChainTrustVerifier.IsTrusted(string assemblyPath, X509Certificate2 signerCertificate)` never receives the `SignedCms`, so the intermediates are not even reachable from here. The `SignedCms` (or an extra certificate collection) has to be passed through the interface.

src/PluginSystem/SAF.PluginSystem.Hosting.Extensions/Authenticode/CrossPlatformAuthenticodeTrustVerifier.cs
```diff
@@ -0,0 +1,38 @@
...
+    // 1.3.6.1.5.5.7.3.3 - id-kp-codeSigning
+    private const string CodeSigningEnhancedKeyUsageOid = "1.3.6.1.5.5.7.3.3";
+
+    // This verifier only validates the certificate chain; the file binding must be checked
+    // separately, so callers must not treat a trusted result as proof the signature covers the file.
+    public bool VerifiesFileIntegrity => false;
+
+    public bool IsTrusted(string assemblyPath, X509Certificate2 signerCertificate)
+    {
+        using var chain = new X509Chain();
```
berndbr: **No Authenticode timestamp / counter-signature handling: a plugin signed with a since-expired certificate is rejected on non-Windows even though its timestamped signature is still valid.**

`chain.Build(signerCertificate)` validates the certificate against *now*. The point of Authenticode is that an RFC3161 / PKCS#9 counter-signature pins the signing time, so `X509ChainPolicy.VerificationTime` should be set from the timestamp in the CMS unsigned attributes (falling back to rejecting when no timestamp is present, if that is the intended policy).

Code-signing certificates are issued for 1-3 years while deployed plugin payloads live much longer, so this fires in normal operation. Verified: building the chain for the `kernel32.dll` leaf yields `NotTimeValid` even once the intermediates are supplied.

`WindowsAuthenticodeTrustVerifier` gets this right via `WinVerifyTrust`, so Windows and non-Windows hosts will disagree about the same file. That is a second semantic difference hidden behind the single `VerifiesFileIntegrity` capability flag.

src/PluginSystem/SAF.PluginSystem.Hosting/PluginAssemblyFolderContainer.cs
```diff
@@ -86,20 +90,33 @@ private List<IPluginManifest> LoadManifests(List<string> pluginAssemblyPaths)
 
         foreach (var pluginAssemblyPath in pluginAssemblyPaths)
         {
-            _logger.LogDebug("Create AssemblyLoadContext for {PluginAssemblyPath}", pluginAssemblyPath);
+            try
+            {
+                using var assemblyFile = _fileSystem.FileInfo.New(pluginAssemblyPath)
```
berndbr, line 102: **New uncaught exception surface: a single locked or ACL-protected DLL in the search root now crashes host startup instead of being skipped.**

Opening the file explicitly introduces exception types that the `catch` filter on line 131 does not list (`BadImageFormatException or FileLoadException or FileNotFoundException or ReflectionTypeLoadException or TypeLoadException`):

- sharing violation (installer, AV scanner, backup agent holding the DLL) produces `IOException`
- restrictive ACLs, or a directory where a file was expected, produces `UnauthorizedAccessException`

Both escape `LoadManifests` into `GetPluginManifests` and abort host startup.

Verified against the same exclusively locked file:

    LoadFromAssemblyPath -> System.IO.FileLoadException   (previous behaviour: caught, logged, skipped)
    FileInfo.Open        -> System.IO.IOException         (current behaviour: escapes)

`AuthenticodeSignatureReader` has the matching gap: `File.Delete(trustSnapshotPath)` in its `finally` can throw `UnauthorizedAccessException`, which is not in its exception filter either.

Adding `IOException or UnauthorizedAccessException` to the filter on line 131 restores the "skip the bad assembly, keep the host alive" contract the rest of the method follows.

src/PluginSystem/SAF.PluginSystem.Hosting/SAF.PluginSystem.Hosting.csproj
```diff
@@ -21,11 +21,13 @@ SPDX-License-Identifier: MPL-2.0
   </ItemGroup>
 
   <ItemGroup>
+    <ProjectReference Include="..\SAF.PluginSystem.Hosting.Extensions\SAF.PluginSystem.Hosting.Extensions.csproj" />
```
berndbr: **This reference inverts the package layering and turns the opt-in validation feature into a mandatory dependency of the core engine.**

No production source file in `SAF.PluginSystem.Hosting` references anything from `SAF.PluginSystem.Hosting.Extensions` (the only hits are the generated `InternalsVisibleTo` under `obj/`). `SAF.PluginSystem.Hosting.Tests.csproj` already has its own direct `ProjectReference` to Extensions, so it is not needed for the tests either.

The cost is real, not theoretical - `dotnet pack` on this branch emits:

    <dependency id="SAF.PluginSystem.Hosting.Contracts" version="11.0.0-alpha.0.48" />
    <dependency id="SAF.PluginSystem.Hosting.Extensions" version="11.0.0-alpha.0.48" />

So every consumer of the core plugin engine now also pulls `SAF.PluginSystem.Hosting.Extensions` and, transitively, `System.Security.Cryptography.Pkcs`. That contradicts `docs/plugin-system.md`:

> Plugin assembly validation is opt-in. SAF does not enable validators by default.

Extensions only references Contracts, so removing this line should be a no-op for the build.

src/PluginSystem/SAF.PluginSystem.Hosting/PluginAssemblyFolderContainer.cs
```diff
@@ -86,20 +90,33 @@ private List<IPluginManifest> LoadManifests(List<string> pluginAssemblyPaths)
...
+                    _fileSystem.Path.GetDirectoryName(AppContext.BaseDirectory),
+                    _fileSystem.Path.GetDirectoryName(pluginAssemblyPath),
+                    StringComparison.OrdinalIgnoreCase) == 0;
+
+                var pluginLoadContext = isInBaseDirectory
+                    ? AssemblyLoadContext.Default
+                    : new PluginAssemblyLoadContext(loggerFactory, pluginAssemblyPath, _fileSystem);
+
+                using var assemblyStream = new MemoryStream(assemblyBytes, writable: false);
+                var assembly = pluginLoadContext.LoadFromStream(assemblyStream);
```
berndbr: **Switching to `LoadFromStream` empties `Assembly.Location`, silently blanking plugin version and build-date diagnostics across the whole host.**

An assembly loaded from a stream has `Location == ""` (verified: loading a DLL into a fresh `AssemblyLoadContext` via `LoadFromStream` yields `Location=''`, `IsDynamic=False`). That is the normal path for every plugin outside the base directory.

Affected call sites, all guarded by `if (!string.IsNullOrEmpty(assembly.Location))` so they degrade silently rather than throwing:

- `src/Hosting/SAF.Hosting/Diagnostics/SafServiceInfo.cs:20` - constructed from `IPluginManifest`, i.e. exactly these assemblies. `Version` and `BuildDate` are now always empty/`default` for every plugin surfaced through `SafNodeInfo.SafServices`.
- `src/Hosting/SAF.Hosting/Diagnostics/SafVersionInfo.cs:18`
- `src/Messaging/SAF.Messaging.Cde/Diagnostics/CdeNodeInfo.cs:36` - returns `null`.
- This file, lines 124 and 128 - `{AssemblyLocation}` now logs an empty string, removing the main clue for diagnosing discovery problems.

Third-party plugin code that resolves resources relative to `Assembly.GetExecutingAssembly().Location` breaks the same way.

If the stream load is kept, these call sites need the discovered path threaded through (the container already has `pluginAssemblyPath`), and the behaviour change belongs in `docs/migration-10-to-11.md`.

src/PluginSystem/SAF.PluginSystem.Hosting/PluginAssemblyFolderContainer.cs
```diff
@@ -86,20 +90,33 @@ private List<IPluginManifest> LoadManifests(List<string> pluginAssemblyPaths)
...
 
-            try
-            {
-                var assembly = pluginLoadContext.LoadFromAssemblyPath(pluginAssemblyPath);
+                var isInBaseDirectory = string.Compare(
+                    _fileSystem.Path.GetDirectoryName(AppContext.BaseDirectory),
+                    _fileSystem.Path.GetDirectoryName(pluginAssemblyPath),
+                    StringComparison.OrdinalIgnoreCase) == 0;
+
+                var pluginLoadContext = isInBaseDirectory
```
berndbr, line 125: **For plugins in `AppContext.BaseDirectory` the "validate the exact bytes you load" guarantee silently does not hold - `AssemblyLoadContext.Default.LoadFromStream` can return an assembly loaded from disk instead of the validated snapshot.**

When `isInBaseDirectory` is true the load goes to `AssemblyLoadContext.Default`, whose binder resolves by identity first. Verified:

- assembly already loaded in `Default`: `LoadFromStream` returns the existing instance (`ReferenceEquals == true`, real `Location`, still one entry in `Default.Assemblies`); the snapshot bytes are discarded.
- identity present in the app TPA list: `LoadFromStream` returns the framework/app copy with its on-disk `Location`.
- identity unknown to `Default`: the snapshot bytes are used (`Location == ""`).

`docs/getting-started.md` and `docs/plugin-system.md` both recommend `options.SearchRootPath = AppContext.BaseDirectory`, and `SAF.Messaging.Runtime.dll` is described as added automatically by `AddSafHost`, i.e. referenced by the host and therefore in TPA. So the documented common configuration lands on the branch where validation and loading are decoupled again.

`docs/plugin-system.md:391` states this as an unconditional guarantee:

> The hosting pipeline reads each candidate once into a content snapshot, validates that snapshot, and loads the same bytes with `LoadFromStream`.

Either qualify the claim or route base-directory candidates through a dedicated load context too.

src/PluginSystem/SAF.PluginSystem.Hosting/PluginAssemblyFolderContainer.cs
```diff
@@ -119,4 +136,68 @@ private List<IPluginManifest> LoadManifests(List<string> pluginAssemblyPaths)
...
+        {
+            rejectionReason = $"metadata could not be read ({ex.GetType().Name})";
+            return false;
+        }
+
+        var validationContext = new PluginAssemblyValidationContext(pluginAssemblyPath, assemblyName, assemblyBytes);
+
+        foreach (var validator in _assemblyValidators)
+        {
+            var result = validator.Validate(validationContext);
```
berndbr: **A third-party `IPluginAssemblyValidator` that throws takes down host startup.**

`validator.Validate(validationContext)` is called with no guard. `IPluginAssemblyValidator` is a public extension point in `SAF.PluginSystem.Hosting.Contracts` and `docs/plugin-system.md` explicitly invites users to implement it, so an `InvalidOperationException`, `ArgumentException`, `CryptographicException` or anything else thrown by someone else's validator propagates out of `TryValidateAssembly` into `LoadManifests` and `GetPluginManifests`, and aborts the host.

The rest of this method is carefully written to fail closed per assembly (`rejectionReason` plus `continue`). The validator call should follow the same contract - catch, log, and treat the assembly as rejected - so one broken validator cannot deny service for the entire host.

src/PluginSystem/SAF.PluginSystem.Hosting.Extensions/DigitalSignaturePluginAssemblyValidator.cs
```diff
@@ -0,0 +1,68 @@
...
+        ArgumentNullException.ThrowIfNull(context);
+
+        var options = _optionsMonitor.Get(_optionsName);
+        var signature = context.AssemblyBytes.IsEmpty
+            ? _authenticodeSignatureReader.ReadSignature(context.AssemblyPath)
+            : _authenticodeSignatureReader.ReadSignature(context.AssemblyBytes);
+        var signerThumbprint = signature?.SignerThumbprint;
+        var hasValidDigitalSignature = signature?.HasValidDigitalSignature ?? false;
+
+        if (options.RequireValidDigitalSignature && !hasValidDigitalSignature)
```
berndbr, line 64: **`AddDigitalSignaturePluginAssemblyValidator()` with no configuration is a complete no-op that accepts unsigned assemblies - a dangerous default for a security control.**

With `RequireValidDigitalSignature == false` and an empty `AllowedSignerThumbprints`, both guards are skipped and every candidate - including a completely unsigned DLL - reaches `PluginAssemblyValidationResult.Accepted()`. Verified: an unsigned assembly validated through default-constructed options is accepted.

The test suite encodes this as intended behaviour (`PluginSystemHostBuilderExtensionsTests.AddDigitalSignaturePluginAssemblyValidator_MultipleRegistrations_ShouldKeepOptionsIsolated` asserts `validators[0].Validate(context).IsAccepted` is `true` for an unsigned assembly), and `docs/plugin-security.md` presents registering the validator as "an additional check for discovered entry assemblies".

A registered-but-unconfigured validator that silently allows everything is worse than no validator, because it reads as protection in the composition root. Consider defaulting `RequireValidDigitalSignature` to `true`, or adding an `IValidateOptions<T>` that fails fast when neither check is configured.

src/PluginSystem/SAF.PluginSystem.Hosting.Extensions.Tests/Authenticode/AuthenticodeSignatureReaderIntegrationTests.cs
```diff
@@ -0,0 +1,242 @@
...
+using System.Reflection.PortableExecutable;
+using System.Security.Cryptography.Pkcs;
+using System.Security.Cryptography.X509Certificates;
+
+public sealed class AuthenticodeSignatureReaderIntegrationTests
+{
+    private const string SignedAssemblyEnvironmentVariable = "SAF_AUTHENTICODE_SIGNED_ASSEMBLY";
+
+    [Fact]
+    public void ReadSignature_ValidatesTrustedDotNetRuntimeAssembly_UsingWindowsTrust()
```
berndbr, line 16: **Coverage gap that let the production path ship broken: nothing tests "trusted signature via the byte-snapshot overload".**

The matrix as it stands:

| test | overload | expectation |
|---|---|---|
| `ReadSignature_ValidatesTrustedDotNetRuntimeAssembly_UsingWindowsTrust` | `ReadSignature(string)` | trusted |
| `ReadSignature_RejectsUntrustedSigntoolAssembly_UsingWindowsTrust` | `ReadSignature(string)` | untrusted |
| `AuthenticodeSignatureReaderTests.ReadSignature_ReturnsNull_WhenContentSnapshotIsNotSigned` | bytes | unsigned |
| `DigitalSignaturePluginAssemblyValidatorTests` | n/a | `IAuthenticodeSignatureReader` is mocked |

`PluginAssemblyFolderContainer` only ever uses the bytes overload, and the only bytes-overload assertions cover unsigned input or an intentionally untrusted fixture that is expected to return `false`. That is precisely why the `OpenTrustSnapshot` sharing-violation defect passes CI green.

Add a Windows test asserting `HasValidDigitalSignature == true` for a trusted signed file through `ReadSignature(ReadOnlyMemory<byte>)`; the existing `FindTrustedDotNetRuntimeAssembly` helper already locates a suitable file.

Separately: every test in this class starts with `Assert.SkipWhen(signedAssemblyPath is null, ...)`, so on a normal `dotnet test` run without `SAF_AUTHENTICODE_SIGNED_ASSEMBLY` the whole file contributes zero assertions and still reports green.

src/PluginSystem/SAF.PluginSystem.Hosting/PluginAssemblyFolderContainer.cs
```diff
@@ -18,12 +20,14 @@ public class PluginAssemblyFolderContainer(
     ILoggerFactory loggerFactory,
     IPluginManifestLoader manifestLoader,
     PluginAssemblyFolderSearchOptions options,
+    IEnumerable<IPluginAssemblyValidator> assemblyValidators,
```
berndbr: **Undocumented binary and source breaking change to a public type in a shipped package.**

`PluginAssemblyFolderContainer` is `public` and ships in the `SAF.PluginSystem.Hosting` NuGet package. The new required `IEnumerable<IPluginAssemblyValidator>` parameter is inserted in the middle of the parameter list, before `IFileSystem fileSystem`, so every external call site breaks at compile time and any already-compiled caller breaks at runtime.

`docs/migration-10-to-11.md` documents considerably smaller changes (Step 3a is about swapping a file-system abstraction) but says nothing about this one, even though this PR did add a section there for the new `System.Security.Cryptography.Pkcs` reference.

Either add an overload that defaults the validators to `[]`, or append the parameter at the end - and document it under "Step 8 - Package Reference Changes".

docs/plugin-system.md
```diff
@@ -335,6 +338,90 @@ Services are **not** forwarded into plugin containers automatically. Use `IHostS
...
+
+// Optional Authenticode validation
+pluginSystemBuilder.AddDigitalSignaturePluginAssemblyValidator(options =>
+{
+    options.RequireValidDigitalSignature = true;
+    options.AllowedSignerThumbprints.Add("AABBCCDDEEFF00112233445566778899AABBCCDD");
+});
+```
+
+The hosting pipeline reads each candidate once into a content snapshot, validates that snapshot, and loads the same bytes with `LoadFromStream`. The digital-signature validator reads Authenticode signatures from that snapshot (PE certificate table) and validates signer trust using `X509Chain`.
```
berndbr: **Inaccurate: on Windows signer trust is not validated with `X509Chain`.**

> The digital-signature validator reads Authenticode signatures from that snapshot (PE certificate table) and validates signer trust using `X509Chain`.

`AuthenticodeSignatureReader`'s parameterless constructor selects `WindowsAuthenticodeTrustVerifier` on Windows, which calls `WinVerifyTrust` with `WINTRUST_ACTION_GENERIC_VERIFY_V2`. `X509Chain` (`CrossPlatformAuthenticodeTrustVerifier`) is only the Linux/macOS fallback, and it deliberately reports `VerifiesFileIntegrity => false`.

This matters to readers because the two paths have materially different semantics: `WinVerifyTrust` honours the Authenticode timestamp and the Windows trust policy, while the `X509Chain` fallback does neither. Suggest naming both, e.g. "validates signer trust with `WinVerifyTrust` on Windows and with `X509Chain` against the platform certificate store elsewhere".

src/PluginSystem/SAF.PluginSystem.Hosting.Extensions/PluginSystemHostBuilderExtensions.cs
```diff
@@ -0,0 +1,94 @@
...
+        Action<DigitalSignaturePluginAssemblyValidatorOptions>? configure = null)
+    {
+        ArgumentNullException.ThrowIfNull(hostBuilder);
+
+        hostBuilder.Services.AddOptions();
+        var uniqueOptionsKey = $"{nameof(DigitalSignaturePluginAssemblyValidatorOptions)}_{Guid.NewGuid():N}";
+        hostBuilder.Services.Configure<DigitalSignaturePluginAssemblyValidatorOptions>(
+            uniqueOptionsKey,
+            options => configure?.Invoke(options));
+        hostBuilder.Services.AddSingleton<IAuthenticodeCertificateTableParser, AuthenticodeCertificateTableParser>();
```
berndbr: **These four infrastructure singletons are re-registered on every call, so calling the helper twice duplicates the whole Authenticode object graph.**

`AddSingleton` appends unconditionally. `PluginSystemHostBuilderExtensionsTests.AddDigitalSignaturePluginAssemblyValidator_MultipleRegistrations_ShouldKeepOptionsIsolated` already calls this method twice, which is a legitimate scenario (two trust domains, two thumbprint allow-lists). The result is duplicate `IAuthenticodeCertificateTableParser`, `IAuthenticodeChainTrustVerifier`, `IAuthenticodePeHasher` and `IAuthenticodeSignatureReader` registrations. It happens to work because `GetRequiredService` takes the last one, but the descriptors accumulate and anything resolving `IEnumerable<IAuthenticodeSignatureReader>` gets surprises.

`TryAddSingleton` (from `Microsoft.Extensions.DependencyInjection.Extensions`) expresses the intent: infrastructure registered once, validator registered per call. The validator registration further down must stay `AddSingleton`.

Two further notes on this file:

- The file is named `PluginSystemHostBuilderExtensions.cs` but declares `PluginAssemblyValidationBuilderExtensions`, which also reads as a duplicate of `SAF.PluginSystem.Hosting.PluginSystemHostBuilderExtensions`.
- `DigitalSignaturePluginAssemblyValidator` and `AuthenticodeSignatureReader` both build the same graph with `new` in their public constructors, so there are two independent composition roots for one dependency graph and only this one is exercised via DI.

src/PluginSystem/SAF.PluginSystem.Hosting/PluginAssemblyFolderContainer.cs
```diff
@@ -86,20 +90,33 @@ private List<IPluginManifest> LoadManifests(List<string> pluginAssemblyPaths)
...
         foreach (var pluginAssemblyPath in pluginAssemblyPaths)
         {
-            _logger.LogDebug("Create AssemblyLoadContext for {PluginAssemblyPath}", pluginAssemblyPath);
+            try
+            {
+                using var assemblyFile = _fileSystem.FileInfo.New(pluginAssemblyPath)
+                    .Open(FileMode.Open, FileAccess.Read, FileShare.Read);
+                using var assemblyBuffer = new MemoryStream();
+                assemblyFile.CopyTo(assemblyBuffer);
+                var assemblyBytes = assemblyBuffer.ToArray();
```
berndbr: **Four full-size copies of every plugin assembly, plus a synchronous fsync per assembly, all on the startup path inside `_cacheLock`.**

Per candidate DLL:

1. `assemblyFile.CopyTo(assemblyBuffer)` - copy 1 (the `MemoryStream` internal buffer)
2. `assemblyBuffer.ToArray()` - copy 2
3. `GetAssemblyName(assemblyBytes)` calls `assemblyBytes.ToArray()` - copy 3
4. `new MemoryStream(assemblyBytes, writable: false)` for `LoadFromStream`, after which the runtime makes its own copy

When the digital-signature validator is registered, `AuthenticodeSignatureReader.ReadSignature(ReadOnlyMemory<byte>)` adds another `.ToArray()` in `TryReadAuthenticodeSignedData`, another in `AuthenticodePeHasher.ComputeAuthenticodeHash`, and `OpenTrustSnapshot` writes the whole payload to a temp file with `Flush(flushToDisk: true)` - a real fsync per assembly on every host start. On non-Windows that temp file and fsync are pure waste, because `CrossPlatformAuthenticodeTrustVerifier.IsTrusted` ignores the `assemblyPath` argument entirely.

Multi-MB plugin assemblies land on the LOH; the previous `LoadFromAssemblyPath` let the runtime memory-map the file instead.

Cheap improvements: reuse one `byte[]` (`MemoryStream.GetBuffer()`, or a single read via `RandomAccess`), pass the same `MemoryStream` with `Position = 0` to `GetAssemblyName`, hoist the `ArrayPool` rental in `AuthenticodePeHasher.HashRange` out of the per-range calls, and skip `OpenTrustSnapshot` entirely when the verifier does not consume the path.

docs/migration-10-to-11.md
```diff
@@ -207,7 +207,11 @@ ps.AddPluginAssemblyFolderContainer(options =>
 
 Without `SAF.Messaging.Runtime`, calls to inject `IMessagingInfrastructure` will throw `InvalidOperationException` at runtime.
 
----
+The `SAF.PluginSystem.Hosting.Extensions` package now references `System.Security.Cryptography.Pkcs` for Authenticode CMS signature parsing. Package consumers normally receive it transitively; projects that declare package dependencies explicitly should also restore this package:
```
berndbr: **This paragraph replaced the `---` separator that every other step boundary in the document has, and it is filed under the wrong step.**

`## Step 1` through `## Step 4`, and `## Step 6`, `## Step 7`, `## Step 8`, are each preceded by a `---` horizontal rule. After this change `## Step 5 - Migrate Plugin Lifecycle Code` is the only one that is not, so the rendered page runs Step 4 straight into Step 5.

Content-wise this note is about package references, so it belongs under `## Step 8 - Package Reference Changes` rather than at the end of the `SAF.Messaging.Runtime` discoverability step.

Also worth reconsidering the claim "Package consumers normally receive it transitively": on this branch `SAF.PluginSystem.Hosting` itself takes a hard dependency on `SAF.PluginSystem.Hosting.Extensions`, so consumers receive `System.Security.Cryptography.Pkcs` whether or not they use assembly validation.

src/PluginSystem/SAF.PluginSystem.Hosting/SAF.PluginSystem.Hosting.csproj
berndbr: Fixed in 96b19a4: the `ProjectReference` is removed. Verified with `dotnet pack` — the nuspec dependency group now contains only `SAF.PluginSystem.Hosting.Contracts` plus the `Microsoft.Extensions.*` / `Testably.Abstractions` packages, so `SAF.PluginSystem.Hosting.Extensions` and its transitive `System.Security.Cryptography.Pkcs` no longer ship as dependencies of the core engine. Solution builds clean and `SAF.PluginSystem.Hosting.Tests` passes (146/146). The migration-guide claim that consumers receive the package transitively will be corrected in a follow-up commit.

src/PluginSystem/SAF.PluginSystem.Hosting/PluginAssemblyFolderContainer.cs
berndbr: Fixed in `deb9d97`.

`IEnumerable<IPluginAssemblyValidator> assemblyValidators` moved to the **end** of the primary constructor, after `IFileSystem fileSystem`. It stays required (an empty collection is the no-validator case, and DI supplies one automatically), but the common positional shape `..., options, fileSystem)` is the same as in 10.x again — only callers that explicitly passed validators are affected.

Call sites updated: the DI factory in `PluginSystemHostBuilderExtensions`, `PluginAssemblyFolderContainerTests` and `PluginServicesIsolationTests`. The break gets documented under `## Step 8 - Package Reference Changes` in `docs/migration-10-to-11.md` together with the other migration notes from this review.

src/PluginSystem/SAF.PluginSystem.Hosting/PluginAssemblyFolderContainer.cs
berndbr: Fixed in `deb9d97`.

The filter now reads:

```csharp
catch (Exception ex) when (ex is BadImageFormatException or FileLoadException or FileNotFoundException
                              or ReflectionTypeLoadException or TypeLoadException
                              or IOException or UnauthorizedAccessException)
```

So a candidate held by an AV scanner, installer or backup agent (sharing violation) and one with restrictive ACLs are logged and skipped, instead of aborting host startup.

Regression tests added, each with two candidates so the assertion also proves the *other* assembly still loads:
- `GetPluginManifests_SkipsAssembly_WhenFileIsExclusivelyLocked` — holds a `FileShare.None` handle across the call.
- `GetPluginManifests_SkipsAssembly_WhenAccessIsDenied` — deny-read ACL, Windows only.

Your follow-up note about `File.Delete(trustSnapshotPath)` in `AuthenticodeSignatureReader` is tracked separately: the snapshot disappears from the production path once the container hands validators a pinned path, and the remaining byte-overload cleanup will swallow-and-log rather than propagate. That lands with the fix for the trust-snapshot handle issue.

src/PluginSystem/SAF.PluginSystem.Hosting/PluginAssemblyFolderContainer.cs
berndbr: Fixed in `deb9d97`.

`validator.Validate(validationContext)` is wrapped now:

```csharp
try
{
    result = validator.Validate(validationContext);
}
catch (Exception ex)
{
    // IPluginAssemblyValidator is a public extension point: a throwing implementation
    // must not abort the host, and the assembly is rejected (fail closed).
    _logger.LogError(ex, "Validator {ValidatorType} threw while validating {PluginAssemblyPath}", validator.GetType().Name, pluginAssemblyPath);
    rejectionReason = $"validator {validator.GetType().Name} threw {ex.GetType().Name}";
    return false;
}
```

Deliberately a bare `catch (Exception)` — the whole point is that third-party code we invite users to write in `docs/plugin-system.md` can throw anything. The candidate is rejected rather than accepted, matching the fail-closed contract of the rest of the method.

Test `GetPluginManifests_SkipsAssembly_WhenCustomValidationThrows`: a validator that throws for one of two candidates → that assembly is skipped, the other one still loads.

src/PluginSystem/SAF.PluginSystem.Hosting/PluginAssemblyFolderContainer.cs
berndbr: Fixed in `480469c`. You were right that blanking `Assembly.Location` is too high a price — it breaks `SafServiceInfo`, `SafVersionInfo`, `CdeNodeInfo`, the two log statements in this method and any plugin that resolves resources next to its own assembly.

`LoadFromAssemblyPath(pluginAssemblyPath)` is back. The validate-then-load coupling is kept per platform instead:

- **Windows**: the `FileShare.Read` handle used to read the candidate is now held across validation *and* the load. That share mode denies every subsequent open asking for write **or delete**, so neither an in-place edit nor a rename/replace swap can happen in between.
- **Other platforms**: file locks are advisory, so a held descriptor cannot keep the *path* stable. The file is compared byte-wise against the validated snapshot immediately before the load, and the candidate is skipped on a mismatch.

The second one shortens the window rather than closing it — from the whole validation (certificate chain building, possibly with network revocation checks) down to the load call. That is deliberate and will be stated plainly in `docs/plugin-system.md`, not oversold.

For context on why we did not do better: an intermediate implementation loaded the validated bytes from a private 0700 staging directory, which does close the window for the plugin assembly. It was dropped, because it protects one file out of many — `AssemblyDependencyResolver` still loads the plugin dependencies by path from the deployment folder, so an attacker who can swap the plugin can swap `Foo.Dependency.dll` next to it — and because it replaces an empty `Assembly.Location` with a *wrong* one: the staging directory holds no satellite assemblies and no config files, so resource resolution fails silently. It also added a temp-path dependency and a cleanup lifecycle.

The re-comparison is skipped where it buys nothing: on Windows (the pin is stronger) and when no validator is registered (no validation result a swap could invalidate). Candidate paths are also made absolute now, because `LoadFromAssemblyPath` rejects the relative paths a relative `SearchRootPath` produced.

Tests: `MatchesFileContent` covered for equal / one-bit-changed / shorter / longer content across a multi-chunk buffer, plus an end-to-end test where a validator rewrites the file during validation — the write is denied on Windows and the candidate is skipped elsewhere.

src/PluginSystem/SAF.PluginSystem.Hosting/PluginAssemblyFolderContainer.cs
berndbr: Addressed in `480469c`, see the detailed reply on the `LoadFromStream` thread.

On your point about the base-directory branch specifically: the claim that the validated bytes are the ones that get loaded cannot be upheld here, on any platform. These candidates go to `AssemblyLoadContext.Default`, whose binder resolves by **identity** first, so an assembly of the same identity that is already loaded — or that is on the application TPA list, which is how `SAF.Messaging.Runtime.dll` gets there via `AddSafHost` — wins over the file we just validated. Neither the pinning handle nor a re-comparison can override that.

So instead of pretending otherwise, base-directory candidates are loaded from their original path with no extra ceremony, and the limitation gets written down explicitly in `docs/plugin-system.md:391`. That documentation rewrite is the remaining piece of this thread and is tracked as its own step, together with the load-mechanics paragraph.

src/PluginSystem/SAF.PluginSystem.Hosting.Extensions/Authenticode/AuthenticodeSignatureReader.cs
berndbr: Fixed in `70fc647`, and you were right about the consequence: with `RequireValidDigitalSignature = true` this rejected **every** plugin on Windows.

The mechanism is a share-mode conflict. `WinVerifyTrust` opens the file without sharing writers, so while our `FileAccess.ReadWrite` handle was alive its own open failed with a sharing violation — which the caller cannot distinguish from an untrusted signature, so `IsTrusted` was always `false`.

Rather than repairing the handle, the temporary file is gone entirely. An Authenticode signature does not need to be on disk to be verified:

- signature math — `SignedCms.Decode` + `CheckSignature`
- file binding — `AuthenticodePeHasher` recomputes the Authenticode PE hash over the buffer, which never needed a file
- trust anchor — `X509Chain`

`WinVerifyTrust` was the only file-bound piece, and it adds nothing those three do not cover. So signature checking no longer depends on a writable temp directory at all, which also removes the cleanup path and with it your `File.Delete` follow-up from the other thread.

`ReadSignature(string)` still uses `WinVerifyTrust` — and after the load-path change that is the route the plugin pipeline actually takes, since the container always supplies an absolute path. The in-memory route uses the same chain engine and the same certificate stores; what it does not apply is the Authenticode policy layer above the chain, notably the weak-hash policy. That is documented on the interface.

I verified the fix rather than reasoning about it: with the old write-capable handle restored, the new regression test fails; with the fix it passes. The test asserts `HasValidDigitalSignature == true` for a trusted signed runtime assembly through the snapshot overload — the assertion whose absence let this ship.

src/PluginSystem/SAF.PluginSystem.Hosting.Extensions/Authenticode/CrossPlatformAuthenticodeTrustVerifier.cs
berndbr: Fixed in `70fc647`.

`chain.ChainPolicy.ExtraStore.AddRange(signedCms.Certificates)` is in place. As you noted, Authenticode carries the leaf and its intermediates inside the PKCS#7 blob while platform certificate stores hold roots only, so without this the chain ends in `PartialChain` on every host that cannot fetch the issuer via AIA — the normal case for the air-gapped and egress-restricted machines this framework runs on.

This needed an interface change first: `IAuthenticodeChainTrustVerifier.IsTrusted(string, X509Certificate2)` never received the `SignedCms`, so the embedded certificates were not even reachable from the verifier. It is now `IsTrusted(string? assemblyPath, SignedCms signedCms)`, which also made the timestamp fix on the neighbouring thread reachable.

Chain policy construction is split out into `CreateChainPolicy`, for a reason worth recording: `IsTrusted` returns `false` for an untrusted root either way, so a test through the public surface cannot tell whether the extra store had any effect — the bug would stay invisible. The test asserts the policy directly instead, that both the signer and an additional embedded certificate land in `ExtraStore`.

src/PluginSystem/SAF.PluginSystem.Hosting.Extensions/Authenticode/CrossPlatformAuthenticodeTrustVerifier.cs
berndbr: Fixed in `70fc647`, following the decision that the non-Windows verifier should behave like the Windows one: use the counter-signature time when the signature is timestamped, and the current time otherwise. That is what `WinVerifyTrust` does under `WINTRUST_ACTION_GENERIC_VERIFY_V2` with the default policy, so both platforms now agree instead of Linux quietly rejecting intact signatures whose signing certificate has since expired.

`CreateChainPolicy` sets `VerificationTime` from the signature timestamp, looking in two places:

1. the RFC 3161 token in the unsigned attributes (`1.2.840.113549.1.9.16.2.14`), decoded with `Rfc3161TimestampToken.TryDecode`
2. otherwise the legacy PKCS#9 counter-signature signing time (`1.2.840.113549.1.9.5`)

Tested: `VerificationTime` equals the counter-signature time for a counter-signed CMS, and is the current time when the signature is not timestamped.

One gap I could not close, so it is better said than left unnoticed: **the RFC 3161 branch has no test**. The BCL only has the client side (`Rfc3161TimestampRequest`) and no way to issue a token, so covering it needs a real TSA or a checked-in signed-and-timestamped fixture. The PKCS#9 path, which is what Authenticode has traditionally used, is covered. I would fold the fixture into the test-coverage item if you want that branch covered too.

src/PluginSystem/SAF.PluginSystem.Hosting/PluginAssemblyFolderContainer.cs
berndbr: Fixed in 4ae3e75.

The candidate is now read once into an array sized from the file length instead of growing a `MemoryStream` and calling `ToArray()` on it, and `GetAssemblyName` wraps that same array rather than receiving another copy. `RandomAccess` is not reachable here — `IFileSystem` hands out a `FileSystemStream`, not a `SafeFileHandle` — so the read uses `ReadExactly`. The length is guarded against `Array.MaxLength`, since truncating it to `int` would validate less content than the loader afterwards reads.

The Authenticode snapshot overloads each copied the image again just to get a stream over memory they already had. Both now share `SnapshotStream.Create`, which reads array-backed memory in place via `MemoryMarshal.TryGetArray` and only copies memory that exposes no array. `AuthenticodePeHasher` rents its 80 KB buffer once per image instead of once per hashed range.

The temp file and its `Flush(flushToDisk: true)` are gone as of B1 (70fc647), which also settles the `File.Delete` follow-up from comment #4.

Remaining by design: the image itself still lands on the LOH for multi-MB plugins. Both the strong-name check and the PE hash need the complete image, so the buffer stays — what is gone are the copies of it.

src/PluginSystem/SAF.PluginSystem.Hosting.Extensions/PluginSystemHostBuilderExtensions.cs
berndbr: Fixed in 4af49ca — all three points.

**`TryAddSingleton`.** The four Authenticode services are registered once now; the validator stays `AddSingleton` because it is the part bound to this call's options. Beyond that, the registration no longer writes the graph out at all: `IAuthenticodePeHasher` and `IAuthenticodeSignatureReader` are resolved by constructor injection instead of factory lambdas, so the shape is described only by the constructors and cannot fall out of step with them. Only the trust verifier keeps a factory, since it is chosen from the operating system.

`AddDigitalSignaturePluginAssemblyValidator_MultipleRegistrations_ShouldShareTheAuthenticodeServices` asserts the descriptor counts directly: two `IPluginAssemblyValidator`, one of each Authenticode service.

**File name.** Renamed to `PluginAssemblyValidationBuilderExtensions.cs`, and the test file and class with it — that was the other half of the shadowing against `SAF.PluginSystem.Hosting.PluginSystemHostBuilderExtensions`.

**Two composition roots.** Removed. `AuthenticodeSignatureReader` had five constructors, four of which assembled their own collaborators; `AuthenticodePeHasher` had two. Each keeps exactly one over its dependencies. `DigitalSignaturePluginAssemblyValidator`'s public constructor is gone as well — its dependency is internal, so the surviving constructor is internal too and the registration is now the only way to build one. That constructor was introduced by this PR and has never shipped, so nothing released breaks; the removal is noted for the migration guide.

The tests were the remaining copy of the graph — twelve `new AuthenticodeSignatureReader()` call sites. They go through `AuthenticodeReaderFactory.CreateDefault()`, which resolves the reader **from the production registration**, so the default graph including the platform-dependent verifier is now asserted rather than re-declared. `Create(verifier)` covers the one collaborator DI cannot supply.

One sharp edge left deliberately: `AddPluginAssemblyValidator<DigitalSignaturePluginAssemblyValidator>()` still compiles but fails at resolve time for lack of a public constructor. Documented on the type rather than blocked with machinery.

src/PluginSystem/SAF.PluginSystem.Hosting.Extensions/DigitalSignaturePluginAssemblyValidator.cs
berndbr: Fixed in 8e8d4ce — both remedies, as agreed.

**Default.** `RequireValidDigitalSignature` defaults to `true`, so `AddDigitalSignaturePluginAssemblyValidator()` without configuration demands a signature that is intact, covers the file and chains to a trusted root.

**Fail-fast.** `DigitalSignaturePluginAssemblyValidatorOptionsValidator` refuses the combination you describe — requirement off *and* the allow-list empty — registered with `TryAddEnumerable` so repeated registrations add it once. Note the remaining lenient shape is not a hole: requirement off with a non-empty allow-list still demands a signature that covers the file, it only skips the trust chain.

One thing the `IValidateOptions<>` could not do on its own. Each registration configures its own *named* options, and named options are validated on the first `Get` — which is inside `Validate`, i.e. on the first plugin checked, well after the validator has been reported as present. So the validator reads its options once in its constructor and discards them, which makes *resolving* it the thing that fails. `ValidateOnStart` would express this better, but it lives in `Microsoft.Extensions.Hosting` and this package must not take a dependency on the hosting stack (see #5). Covered by `AddDigitalSignaturePluginAssemblyValidator_WithEveryCheckSwitchedOff_ShouldFailWhileComposing`.

**The test that encoded the old behaviour.** `Validate_Accepts_WhenNoSignatureChecksAreConfigured` is now `Validate_Rejects_WhenNothingIsConfigured`. `AddDigitalSignaturePluginAssemblyValidator_MultipleRegistrations_ShouldKeepOptionsIsolated` no longer needs an accepted unsigned assembly to prove its point: it asserts both registrations reject the same assembly for *different* reasons, which shows the options are isolated without also claiming that checking nothing is acceptable.

One further test needed a real change rather than a renamed assertion: `Validate_Rejects_WhenAllowListIsConfiguredButSignatureDoesNotCoverFile` was only reaching the allow-list branch because the requirement used to default off. It now switches it off explicitly, so it still exercises the branch its name claims.

`docs/plugin-security.md` states the new default and the refused combination. The migration guide entry comes with the documentation pass (#15).

src/PluginSystem/SAF.PluginSystem.Hosting.Extensions.Tests/Authenticode/AuthenticodeSignatureReaderIntegrationTests.cs
berndbr: Closed in 9377dd9, together with the cross-OS work that landed earlier in this PR.

**The matrix gap itself** was filled by `AuthenticodeSignedImageTests`: `AuthenticodeTestPeSigner` builds a real signed PE in memory, so a *trusted* verdict is now asserted through **both** overloads — `ReadSignature(string)` and `ReadSignature(ReadOnlyMemory<byte>)` — on every operating system, for SHA-256/384/512, with no fixture and therefore no skip. Trust is anchored in a throwaway root via `X509ChainTrustMode.CustomRootTrust`, which is the only portable way to obtain a trusted chain without touching machine state. `AuthenticodeTrustedRootIntegrationTests` repeats the same two assertions against the real platform trust stores, in a CI job that installs the fixture certificate into `LocalMachine\Root` (Windows) or via `update-ca-certificates` (Linux) and runs separately from the job expecting that same fixture to be untrusted.

**The point about skips hiding regressions was the sharper one, and it is what 9377dd9 is about.** Every fixture-dependent test began with `Assert.SkipWhen(...)`, so a green run proved nothing — which is precisely how the snapshot route shipped broken. The lookups moved into `AuthenticodeFixtures`, which decides skip-or-fail next to the lookup itself:

- On a developer machine a missing fixture stays a skip — a signtool signature cannot be produced there, and failing would be noise.
- The integration job now sets `SAF_AUTHENTICODE_REQUIRE_FIXTURE=1`, which turns every missing fixture into a *failing test*. The trusted-root job needs no second flag: `SAF_AUTHENTICODE_TRUSTED_ROOT=1` already asserts that the job installed the root, so a missing fixture fails there too.

A renamed variable or a broken artifact download now surfaces red instead of green-with-nothing-asserted. Verified by running both filters with the flags but without a fixture: 4 and 3 failures respectively, while the genuinely platform-conditional skips (`SkipUnless(IsWindows)`) remain skips.

One assertion was also weaker than it read: `ReadSignature_FromContentSnapshot_AgreesWithTheFileRoute` only compared the two routes, which also holds when *both* report untrusted — exactly the state the regression produced. On Windows the verdict is known, so it is now asserted outright.

Before requiring the runtime-assembly lookup I checked the ubuntu integration job's log: those tests pass on Linux (only the two Windows-only ones skip), so the new requirement cannot break that matrix leg.

docs/plugin-system.md
berndbr: Fixed in 5ffb7a3. The trust claim was wrong, and so was the sentence around it — it still described `LoadFromStream`, which this PR has since replaced with `LoadFromAssemblyPath`.

The correction names both mechanisms and says why the distinction matters: signer trust is decided by `WinVerifyTrust` on Windows and by `X509Chain` against the platform certificate store elsewhere; the cross-platform verifier validates only the certificate chain and leaves file integrity to the recomputed PE hash, whereas `WinVerifyTrust` also applies the Authenticode policy layer above the chain.

The paragraph became five, because the guarantee is not uniform and one sentence could only be right about part of it:

1. Read-validate-load with the `FileShare.Read` handle held across both, so `Assembly.Location` reports the deployment path again.
2. Windows pins the file (deny-write, deny-delete); Linux/macOS re-compare against the validated snapshot immediately before the load — stated plainly as *shortening* the window, not closing it.
3. Neither mechanism reaches the plugin's dependencies, which `AssemblyDependencyResolver` loads from the deployment folder unvalidated.
4. Base-directory candidates go to `AssemblyLoadContext.Default`, whose binder resolves by identity, so an already-loaded or host-shipped assembly of the same identity wins and the validated file is never loaded.
5. The corrected trust claim above.

Two details came from reading the code rather than from the plan I worked to: the re-comparison has **no** base-directory carve-out (it is skipped on Windows and when no validator is registered, nothing else), and `AddSafHost` does not put `SAF.Messaging.Runtime.dll` on the default binder's platform-assembly list — it discovers it from the base directory, where the host's own package reference already placed it. The doc states the latter.

docs/migration-10-to-11.md
berndbr: Fixed in 5ffb7a3. The `---` before `## Step 5` is back, and the `System.Security.Cryptography.Pkcs` paragraph moved out of the messaging-runtime step into `## Step 8 — Package Reference Changes`, where it belongs.

The claim inside it was also wrong and did not survive the move: consumers do *not* receive the package transitively through the core engine. `SAF.PluginSystem.Hosting` does not reference `SAF.PluginSystem.Hosting.Extensions`, so applications that load plug-ins without validating them are not made to carry the validators or their cryptography dependencies. Step 8 now says that assembly validation lives in its own package and must be referenced explicitly, and that `System.Security.Cryptography.Pkcs` comes along transitively *from that package*.

Step 8 gained three further subsections while I was there: the `PluginAssemblyFolderContainer` constructor break (validators appended as the last, required parameter — pass `[]` to keep the old behaviour), the secure-by-default digital-signature validator, and the note that `DigitalSignaturePluginAssemblyValidator` has no public constructor. Plus one checklist entry.

Two items on my own to-do list for this guide were dropped rather than written, because they do not survive a check against `origin/main`:

- **`Assembly.Location`.** `origin/main` already loads with `LoadFromAssemblyPath`; the stream load never left this branch. Nothing was reversed, so for anyone migrating from 10.x the property behaves as it always did.
- **The `RequireValidDigitalSignature` framing.** No `DigitalSignature*` file exists on `origin/main` — the whole validator surface is new in this PR — so "the alphas accepted every assembly" would have been false. It is documented as the default of a new feature, with the warning that unsigned or untrusted plug-ins are skipped once the validator is registered.

A migration guide should not invent history for readers coming from 10.x, which is also why the package-split section states what the layering *is* rather than that Hosting "no longer references" Extensions: that reference was added and removed inside this branch. For the same reason nothing was added to **Summary of Breaking Changes** — every item here is new-in-11 API, and that table documents 10.x → 11.x.

PR #111, opened 2026-08-13
Add secret configuration provider

berndbr:
A MS configuration extension configuration provider that stores/reads secrets in/from Windows Credential Manager or certificate protected file on linux.


src/Configuration/SAF.Configuration.Secrets/Configuration/SecretResolvingConfigurationProvider.cs
```diff
@@ -0,0 +1,179 @@
...
+    private ISecretReader BuildBootstrapReader(
+        Action<SecretStoreOptions>? configure,
+        Action<ISecretStoreBuilder>? configureProviders)
+    {
+        var services = new ServiceCollection();
+        services.AddLogging();
+        var storeBuilder = services.AddSecretStore(configure);
+        if (configureProviders is null)
+        {
+            storeBuilder.AddDefaults();
```
berndbr, line 147: **[Critical] The bootstrap container never sees the host-registered `ISecretProtector`, so `AddSecretConfigurationResolution` hard-fails at startup on Linux (and for any file-store setup).**

`BuildBootstrapReader` creates a fresh, minimal `ServiceCollection` containing only `AddLogging()` + `AddSecretStore(configure)` + `AddDefaults()`. On non-Windows, `AddDefaults()` registers only `FileSecretStore`, and `FileSecretStore.IsAvailable` returns `false` without an `ISecretProtector`. `CompositeSecretStore.SelectProvider()` then throws `InvalidOperationException("No available secret store provider was found...")`.

That exception is raised from `Resolve()` -> `Load()` -> `ConfigurationRoot` ctor -> `PluginSystemHostContext.BuildPluginConfiguration` -> host startup. It is not recoverable.

**Failure scenario:** a Linux host follows `docs/secret-store.md` exactly — `hostBuilder.Services.AddSingleton<ISecretProtector>(_ => new PkcsSecretProtector(certificate));` then `ps.AddSecretConfigurationResolution(o => o.Namespace = "myapp")` — and any plugin config value starts with `secret://`. The protector lives in the *host* container; the bootstrap container has no idea it exists. The host crashes at startup with "No available secret store provider was found. Registered providers: file."

The chain is already proven by two tests in this PR: `FileSecretStoreTests.IsAvailable_False_WhenProtectorMissing` and `FileSecretStoreRegistrationTests.AutoSelection_WithoutProtector_ReportsNoAvailableProvider` (whose own comment says it "Mirrors AddDefaults() on a non-Windows host").

The same class of failure hits Windows too: `configure: o => o.ProviderName = "file"` gives "The secret store provider 'file' is not available in this environment." And a custom `AddProvider<TProvider>()` whose ctor depends on host-registered services fails with an opaque DI resolution error, because the bootstrap container only has `AddLogging()`.

Options: let the caller contribute registrations to the bootstrap container (the `configureProviders` callback exposes `ISecretStoreBuilder.Services`, but nothing documents that this is the *only* place a protector can be registered for bootstrap resolution), or defer resolution until the host container is bound instead of resolving eagerly in `Load()`.

src/Configuration/SAF.Configuration.Secrets/FileStore/FileSecretStore.cs
```diff
@@ -0,0 +1,252 @@
...
+            _fileSystem.Directory.CreateDirectory(directory);
+        }
+
+        document.Protector = Protector.Name;
+        var json = JsonSerializer.Serialize(document, JsonOptions);
+
+        // Write to a sibling temp file first, then replace, so a crash mid-write cannot corrupt the store.
+        var tempPath = path + ".tmp";
+        await _fileSystem.File.WriteAllTextAsync(tempPath, json, cancellationToken).ConfigureAwait(false);
+        _fileSystem.File.Move(tempPath, path, overwrite: true);
```
berndbr: **[High/Security] Every write silently destroys the file ACL / mode the docs delegate to the installer.**

`docs/secret-store.md` states: *"**File permissions are the installer's responsibility.** ... Lock the file down at deployment time so only the service account can read it"* (`0600` on Linux, an NTFS ACL on Windows). But the write path creates a *new* file (`path + ".tmp"`) with default inherited/umask permissions and then replaces the destination with it. The replacement carries the **temp file's** security descriptor, not the destination's.

**Verified empirically on this machine:**
```
# before (installer-hardened, inheritance disabled)
secrets.json AMC\Brachmaier:(R)

# after replace-move
secrets.json <inherited>:(I)(RX,W)
             <inherited>:(I)(RX,W)
             NT-AUTORITAET\SYSTEM:(I)(F)
             VORDEFINIERT\Administratoren:(I)(F)
             AMC\Brachmaier:(I)(F)
```
The explicit hardened ACE is gone and the parent directory's inherited ACEs (including two additional principals with write access) are back. On Linux `rename(2)` behaves the same way: the destination inode is replaced by the source inode, whose mode came from the process umask (typically `0644`), so a `chmod 600` is lost.

**Failure scenario:** an installer provisions `/var/lib/myapp/secrets.json` and chmods it `0600`. The service calls `SetSecretAsync` once (rotation, provisioning tooling, anything). The file is now `0644` — world-readable — and stays that way. The only thing standing between an unprivileged local user and the ciphertext is the certificate, and the doc's threat model explicitly relies on file permissions as a layer.

Also note the `.tmp` file itself is created world-readable and, if the `Move` fails, is left behind on disk outside whatever ACL was applied to `<path>`.

Fix: capture the destination's ACL/mode before the move and re-apply it after (or write in place with `FileMode.Truncate` on an existing file, or create the temp file with the target's permissions up front).

src/Configuration/SAF.Configuration.Secrets/Protection/PkcsSecretProtector.cs
```diff
@@ -0,0 +1,65 @@
...
+
+    /// <inheritdoc />
+    public string Name => ProtectorName;
+
+    /// <inheritdoc />
+    public byte[] Protect(byte[] plaintext)
+    {
+        ArgumentNullException.ThrowIfNull(plaintext);
+
+        var envelope = new EnvelopedCms(new ContentInfo(plaintext), new AlgorithmIdentifier(Aes256Cbc));
```
berndbr, line 56: **[High/Security] No integrity protection: CMS `EnvelopedData` with AES-256-**CBC** is unauthenticated, and the store document as a whole is unsigned.**

`EnvelopedCms` (RFC 5652 §6) provides confidentiality only — there is no MAC or AEAD tag. AES-CBC ciphertext is malleable. Nothing in `FileSecretStore`'s `SecretDocument` binds an entry's ciphertext to its key, to the other entries, or to a version.

**Failure scenario (attacker with write access to the secrets file — which, per the `File.Move` finding, is the default state after the first write):**
- **Swap:** copy the base64 blob of `myapp/db/readonly-password` over `myapp/db/admin-password`. Both decrypt cleanly under the same certificate; the service silently authenticates with the wrong credential and nothing detects it.
- **Replay/rollback:** restore an old blob for a rotated secret. The service happily uses the revoked credential.
- **Delete:** remove an entry. Per the `Resolve()` fail-open behaviour, the bound configuration value becomes `null` rather than an error.
- **Bit-flipping:** CBC malleability lets an attacker flip chosen plaintext bits in block *n* by corrupting block *n-1*; if decryption failures are distinguishable (they surface as different exception types/messages here), a padding oracle is in reach.

The `Protector` name stamped into the JSON is attacker-controlled and only guards against misconfiguration, so it adds no integrity.

Mitigation: sign the serialized document with `SignedCms` (or an HMAC keyed from the certificate) and verify on read, and bind the entry name into the protected payload as associated data so a swap is detectable. At minimum, `docs/secret-store.md`'s "Security model" section should state explicitly that tampering with the store file is **not** detected.

src/Configuration/SAF.Configuration.Secrets/Configuration/SecretResolvingConfigurationProvider.cs
```diff
@@ -0,0 +1,179 @@
...
+    }
+
+    private Dictionary<string, string?> BuildData()
+    {
+        var data = new Dictionary<string, string?>(StringComparer.OrdinalIgnoreCase);
+        foreach (var pair in _inner.AsEnumerable())
+        {
+            if (pair.Value is not null && SecretReference.TryParse(pair.Value, _options.ReferencePrefix, out var reference))
+            {
+                data[pair.Key] = Resolve(reference);
```
berndbr, line 107: **[High/Security] Fail-open: an unresolvable secret reference resolves to `null` and shadows the original value, so services start with an empty credential.**

```csharp
data[pair.Key] = Resolve(reference);
```
When `Resolve` returns `null` (secret not provisioned, typo in the name, wrong `Namespace`), the key is still written into `Data`. `ConfigurationProvider.TryGet` returns `true` for a key whose value is `null`, and `ConfigurationRoot` iterates providers in reverse, so this provider wins over the source that held the raw `secret://...` token. The result is `config["Db:Password"] == null` — not the reference, not an error.

This PR's own test documents it: `SecretResolvingConfigurationTests.UnresolvableReference_BecomesNull`.

**Failure scenario:** an operator mistypes `secret://myapp/db/passwrd`. Startup succeeds. `Bind()` produces `Password = null`, and a client that treats a null/empty password as "no authentication" connects anonymously — or worse, a `ConnectionString` built by concatenation silently drops the credential. There is no diagnostic anywhere: nothing is logged when a reference fails to resolve.

Note the behaviour is also *inconsistent*: an infrastructure failure (no available provider) is a hard startup crash, while a missing secret is a silent `null`. `SecretStoreOptions.RequireSecretReferences` exists and is documented as the fail-fast switch but is never read by any code (`docs/secret-store.md` line 297 admits "The transparent resolver does not enforce it yet").

At minimum, log a warning naming the unresolved reference; ideally offer a fail-closed mode that throws when a reference cannot be resolved.

src/Configuration/SAF.Configuration.Secrets/Configuration/SecretResolvingConfigurationProvider.cs
```diff
@@ -0,0 +1,179 @@
...
+    private string BuildEnvironmentVariableName(string name)
+    {
+        var builder = new StringBuilder(_options.EnvironmentVariablePrefix).Append("__");
+        foreach (var character in name)
+        {
+            if (char.IsLetterOrDigit(character))
+            {
+                builder.Append(character);
+            }
+            else if (character == '/')
```
berndbr, line 167: **[High/Security] The environment-variable override is enabled by default and its name mangling is lossy, so distinct secret names collide onto the same variable.**

`BuildEnvironmentVariableName` maps `/` to `__` and *every other* non-alphanumeric character to a single `_`. That is not injective:

| reference | env var |
|---|---|
| `secret://a/b`  | `SECRET__a__b` |
| `secret://a--b` | `SECRET__a__b` |
| `secret://a-b`  | `SECRET__a_b`  |
| `secret://a.b`  | `SECRET__a_b`  |

**Failure scenario:** a low-trust component's configuration references `secret://tenant--public-key`. An operator (or CI) sets `SECRET__tenant__public_key` for it. That same variable is now returned for `secret://tenant/public-key`, a completely different secret in a different logical path — and because `AllowEnvironmentOverride` defaults to `true` and the env check runs *before* the store, the injected value wins over the OS keystore unconditionally.

Secondary concern, worth documenting even if the default stays: with `AllowEnvironmentOverride = true` by default, anyone who can influence the service's environment block (a systemd drop-in, a container `-e` flag, a compromised sibling process that can spawn the service) can substitute **any** secret without touching the Credential Manager or the encrypted file, defeating the whole point of the store. The docs present this purely as a CI/container convenience and never flag it as an attack surface.

Suggested fixes: escape the mangling injectively (e.g. hex-escape non-alphanumerics, or reject names containing characters outside `[A-Za-z0-9/_-]`), and consider defaulting `AllowEnvironmentOverride` to `false` with opt-in for CI.

src/Configuration/SAF.Configuration.Secrets/Configuration/SecretResolvingConfigurationProvider.cs
```diff
@@ -0,0 +1,179 @@
...
+        _bootstrapReader = new Lazy<ISecretReader>(() => BuildBootstrapReader(configure, configureProviders));
+
+        var innerBuilder = new ConfigurationBuilder();
+        foreach (var source in innerSources)
+        {
+            innerBuilder.Add(source);
+        }
+
+        _inner = innerBuilder.Build();
+        _innerReloadRegistration = ChangeToken.OnChange(() => _inner.GetReloadToken(), Reload);
```
berndbr, line 69: **[Medium-High] An exception thrown during `Reload()` propagates onto a file-watcher/thread-pool thread and terminates the process.**

`ChangeToken.OnChange(() => _inner.GetReloadToken(), Reload)` fires `Reload()` whenever *any* inner source reloads. `Reload()` -> `BuildData()` -> `Resolve()` -> a blocking store round-trip that can throw:

- `JsonException` — `FileSecretStore.ReadDocumentAsync` calls `JsonSerializer.Deserialize` on a truncated/half-written store file with no guard.
- `FormatException` — `Decrypt` calls `Convert.FromBase64String(encoded)` on a corrupted entry.
- `CryptographicException` — CMS decrypt failure (cert rotated, private key ACL changed).
- `InvalidOperationException` — the protector-name mismatch check.
- `Win32Exception` — `CredReadW` returning anything other than `ERROR_NOT_FOUND`.

None is caught. The callback runs under `PhysicalFilesWatcher` -> `CancellationTokenSource.Cancel()`, which rethrows callback exceptions on the `FileSystemWatcher` event thread. An unhandled exception on a thread-pool thread tears down the process.

**Failure scenario:** the service is running normally. An operator edits `plugin-settings.json` (or a deployment touches it) at the same moment the certificate is being rotated. The appsettings reload fires `Reload()`, CMS decrypt throws, and the host dies with a bare `AggregateException` and no correlation to the config change.

Note the plugin system deliberately installs an `OnLoadException` guard on every file source for exactly this failure mode (`PluginSystemHostContext.CreateOnLoadExceptionHandler`: *"so a malformed or half-written file neither crashes host startup nor silently wipes values on reload"*). This provider bypasses that protection. Wrapping `Reload()`'s body in a try/catch that logs and keeps the previous `Data` would match the surrounding design.

src/Configuration/SAF.Configuration.Secrets.Contracts/FileSecretStoreOptions.cs
```diff
@@ -0,0 +1,24 @@
...
+    /// </summary>
+    public string? Path { get; set; }
+
+    /// <summary>
+    /// The principal granted exclusive read access to the store file, e.g. <c>"DOMAIN\\svc-qds"</c>
+    /// or <c>"NT SERVICE\\QDS-2"</c> on Windows (NTFS ACL), or a user/group on Linux (ownership plus
+    /// restrictive permissions). When not set, the file inherits the default protection for the
+    /// configured <see cref="SecretScope"/>.
+    /// </summary>
+    public string? ReaderPrincipal { get; set; }
```
berndbr, line 23: **[Medium-High/Security] `ReaderPrincipal` is fully documented as applying an NTFS ACL / Linux ownership but is never read by any code — it is a security option that does nothing.**

The XML doc promises: *"The principal granted **exclusive read access** to the store file, e.g. `"DOMAIN\\svc-qds"` or `"NT SERVICE\\QDS-2"` on Windows (NTFS ACL), or a user/group on Linux (ownership plus restrictive permissions). When not set, the file inherits the default protection for the configured `SecretScope`."* `SecretScope.ServiceAccount`'s doc reinforces it: *"...combined with a file ACL for `FileSecretStoreOptions.ReaderPrincipal`"*.

A repo-wide grep for `ReaderPrincipal` finds only: this declaration, the two doc references above, and two tests that assert the property round-trips through options binding. `FileSecretStore` never touches `_fileOptions.ReaderPrincipal`, and there is no ACL or `chmod` code anywhere in the feature.

**Failure scenario:** an integrator reads the XML doc, sets `o.ReaderPrincipal = "NT SERVICE\\QDS-2"`, and reasonably concludes the store file is now locked to that account. It is not — the file has whatever permissions it inherited, and no warning is emitted. This is worse than not having the option: it actively creates a false belief about the protection in place.

Note this also contradicts `docs/secret-store.md` line 224, which correctly says the provider *"does **not** set ACLs"*. Either implement it, or remove the option and the `SecretScope` cross-reference until it is implemented (and if it stays, `FileSecretStoreOptions` should say it is currently advisory only).

docs/secret-store.md
```diff
@@ -0,0 +1,323 @@
...
+
+        ps.AddSecretStore(options => options.Namespace = "myapp");
+    });
+```
+
+### 2. Share the contracts assembly across the plug-in boundary
+
+`ISecretStore` lives in its own contracts assembly, which the plug-in system must recognise as a
+**shared contract** so plug-ins resolve the same type the host forwards. Add it to
+`PluginContractsSearchPattern` (SAF's own contracts are added automatically; this one is not):
```
berndbr: **[Medium/Docs] Step 2 names the wrong mechanism — `PluginContractsSearchPattern` has nothing to do with making `ISecretStore` resolvable in a plug-in.**

The doc claims the contracts assembly *"must [be recognised] as a **shared contract** so plug-ins resolve the same type the host forwards"*, and tells the user to add it to `PluginContractsSearchPattern`.

Tracing the actual code:

1. `PluginContractsSearchPattern` is consumed **only** by `PublicServiceTypeRegistry.GetAssemblyNames()`, whose single caller is `PluginServicesContainer.CollectPublicServices` — it decides which services a plug-in **exports** to *other plug-ins*. It plays no part in host->plug-in forwarding.
2. `AddSecretStore` forwards via `HostServiceForwarder<ISecretStore>`, whose `Forward` does `pluginServices.AddSingleton(service)` directly into the plug-in's collection. It never consults the registry.
3. What actually makes the type identity match is `PluginAssemblyLoadContext.IsLoadableInDefaultContext` — an assembly is shared with the default ALC iff a DLL with the same full name sits in the **host's base directory**.

**Failure scenario:** a user deploys `SAF.Configuration.Secrets.Contracts.dll` only alongside their plug-in (not next to the host) and follows this section verbatim. The plug-in ALC loads its own copy, the forwarded singleton's `ISecretStore` type differs from the one the plug-in asks for, and `GetService<ISecretStore>()` returns `null` — with no hint that the search pattern was irrelevant. Conversely, a user who deploys the DLL next to the host does **not** need the setting at all, but is told they do.

The section should say: deploy `SAF.Configuration.Secrets.Contracts.dll` into the host's application directory (that is what makes it shared across plug-in load contexts). Mention `PluginContractsSearchPattern` only if a plug-in needs to export services typed by that assembly.

docs/secret-store.md
```diff
@@ -0,0 +1,323 @@
...
+```
+
+Then reference secrets in the plugin configuration with the `secret://` prefix:
+
+```json
+{
+  "OpcUaConnections": [
+    {
+      "User": "secret://myapp/opcua/conn-1/user",
+      "Password": "secret://myapp/opcua/conn-1/password",
```
berndbr, line 303: **[Medium/Docs] The JSON example double-prefixes the namespace, so copying it yields an unresolvable secret.**

Ten lines above, the same walkthrough sets the namespace:
```csharp
ps.AddSecretConfigurationResolution(o => o.Namespace = "myapp");
```
and then references `"secret://myapp/opcua/conn-1/password"`. But `BuildTargetName` prepends the namespace to the *logical* name:
```csharp
return string.IsNullOrEmpty(ns) ? name : $"{ns}/{name}";
```
So the physical store key becomes `myapp/myapp/opcua/conn-1/password`.

This contradicts the doc's own "Secret names" section (line 303-305): *"A secret name is a logical key such as `opcua/connection-1/password`. The active `Namespace` is prepended to form the physical store key (e.g. `myapp/opcua/connection-1/password`)"*, and the `ISecretStore` usage sample at line 94 (`GetSecretAsync("opcua/connection-1/user", ct)`) which correctly omits the namespace.

**Failure scenario:** the user provisions `myapp/opcua/conn-1/password` (as the "Secret names" section instructs) and copies this JSON. Lookup misses, and per the fail-open behaviour the bound `Password` becomes `null` with no error — a silent, hard-to-diagnose misconfiguration.

The references should read `secret://opcua/conn-1/password` etc. The env-var bullet at line 261 has the same problem (`secret://myapp/opcua/conn-1/password` -> `SECRET__myapp__opcua__conn_1__password`).

src/Configuration/SAF.Configuration.Secrets/Configuration/SecretResolvingConfigurationProvider.cs
```diff
@@ -0,0 +1,179 @@
...
+        if (_accessor is not null)
+        {
+            _accessorReloadRegistration = ChangeToken.OnChange(() => _accessor.GetChangeToken(), Reload);
+        }
+    }
+
+    public override void Load() => Data = BuildData();
```
berndbr, line 74: **[Medium/Security] Every referenced secret is eagerly decrypted into `ConfigurationProvider.Data` as a plaintext `string` for the process lifetime, and is exposed by `IConfigurationRoot.GetDebugView()`.**

`Load()` resolves **all** `secret://` references found anywhere in the merged configuration, not just those a plug-in will actually read, and stores the plaintext in the base `Data` dictionary. Two consequences the docs never mention:

1. **Diagnostics leak.** `ConfigurationRootExtensions.GetDebugView()` walks every provider's `Data` and emits `key=value (ProviderName)`. Any diagnostics path that calls it — a `/config` style endpoint, a support-bundle dumper, a startup dump behind a debug flag — now prints every decrypted secret in clear, attributed to `SecretResolvingConfigurationProvider`. There is a masking overload (`GetDebugView(root, context => ...)`) but nothing here nudges consumers toward it.
2. **Memory hygiene is claimed but not achieved.** `FileSecretStore.Encrypt`/`Decrypt` and `WindowsCredentialManagerNativeApi.ReadBlob`/`WriteGenericCredential` carefully `Array.Clear` their byte buffers, but the very next step is `Encoding.UTF8.GetString(...)`, producing an immutable managed `string` that is never zeroed and survives until GC — and is then retained indefinitely in `Data`. The byte-scrubbing is therefore cosmetic.

**Failure scenario:** an operator hits a diagnostics endpoint that renders `GetDebugView()` to triage a config problem and pastes the output into a support ticket, leaking every credential the host resolved.

This is largely inherent to the `IConfiguration` model, but it is a material weakening of the stated security model (*"the secret is no longer in the file"* — it is now in process memory and in any config dump instead) and belongs in the "Security model" section. Resolving lazily (only on `TryGet`) would at least bound the exposure to secrets actually used.

src/Configuration/SAF.Configuration.Secrets/CompositeSecretStore.cs
```diff
@@ -0,0 +1,81 @@
...
+        ILogger<CompositeSecretStore> logger)
+    {
+        ArgumentNullException.ThrowIfNull(providers);
+        ArgumentNullException.ThrowIfNull(options);
+        ArgumentNullException.ThrowIfNull(logger);
+
+        _providers = [.. providers];
+        _options = options.Value;
+        _logger = logger;
+        _activeProvider = new Lazy<ISecretStoreProvider>(SelectProvider);
```
berndbr, line 36: **[Medium] `Lazy<T>` with the default `ExecutionAndPublication` mode caches the *exception*, permanently latching a transient selection failure for the process lifetime.**

`Lazy<T>` memoizes a factory exception and rethrows the same instance on every subsequent access — it never retries. `CompositeSecretStore` is registered as a **singleton**, so once `SelectProvider()` throws, every `GetSecretAsync`/`SetSecretAsync`/`RemoveSecretAsync` on that store fails forever, even after the underlying condition is fixed.

For the two built-in providers `IsAvailable` is effectively constant, so this is latent today. It stops being latent as soon as anyone uses the documented extension point:

```csharp
ps.AddSecretStore(null, providers => providers
    .AddProvider<MyKeyVaultProvider>()   // 1st priority
    .AddWindowsCredentialManager());     // fallback
```

**Failure scenario:** `MyKeyVaultProvider.IsAvailable` performs a reachability probe and the vault is briefly unreachable while the first secret is resolved. `SelectProvider()` sees no available provider and throws. The vault comes back 5 seconds later — but the cached exception means the store stays dead until the process is restarted, and the fallback Credential Manager is never consulted either.

The same pattern applies to `_bootstrapReader` in `SecretResolvingConfigurationProvider`. Consider re-evaluating selection on failure (or at least documenting on `ISecretStoreProvider.IsAvailable` that it must be a cheap, stable, side-effect-free predicate, since it is queried exactly once per process).

src/Configuration/SAF.Configuration.Secrets/FileStore/FileSecretStore.cs
```diff
@@ -0,0 +1,252 @@
...
+    private readonly IFileSystem _fileSystem;
+    private readonly ISecretProtector? _protector;
+    private readonly SecretStoreOptions _options;
+    private readonly FileSecretStoreOptions _fileOptions;
+    private readonly ILogger<FileSecretStore> _logger;
+
+    // Serializes all file access. Writers must not lose updates in their read-modify-write cycle, and a
+    // read must not overlap the atomic replace of a write: on Windows an open read handle (FileShare.Read,
+    // no delete-share) would make the writer's overwrite Move fail with a sharing violation.
+    private readonly SemaphoreSlim _fileGate = new(1, 1);
```
berndbr, line 51: **[Medium] The `SemaphoreSlim` only serializes *within one process*, but the documented provisioning workflow is a second process — so read-modify-write updates are lost and the fixed `.tmp` name collides.**

The comment claims: *"Serializes all file access. Writers must not lose updates in their read-modify-write cycle, and a read must not overlap the atomic replace of a write."* That holds only for a single `FileSecretStore` instance. Meanwhile `docs/secret-store.md` ("Provisioning secrets") explicitly instructs: *"Use `ISecretStore.SetSecretAsync` from your own tooling/installer"* — i.e. a separate process pointed at the same file.

Three concrete problems across processes:

1. **Lost update.** Installer and service both `ReadDocumentAsync` (full document), each mutates one entry, each `WriteDocumentAsync` replaces the whole file. Last writer wins and silently discards the other's secret. There is no file lock and no optimistic-concurrency stamp.
2. **Temp-file collision.** `var tempPath = path + ".tmp"` is a fixed name. Two concurrent writers race on the same path: one truncates the other's partially written temp file, and the `Move` can publish a half-written document, or fail with a sharing violation on Windows.
3. **Leftover temp file.** If `WriteAllTextAsync` succeeds but `Move` throws, `<path>.tmp` is left on disk holding a full copy of the store, outside any ACL the installer applied to `<path>` (see the `File.Move` finding).

**Failure scenario:** an installer writes `db/password` while the running service writes `db/user` during a rotation. One of the two silently vanishes; on next startup the corresponding config value resolves to `null` and — per the fail-open behaviour — the service starts with an empty credential.

Use a randomized temp name (`path + "." + Guid.NewGuid().ToString("N") + ".tmp"`) and take a cross-process lock (an exclusive `FileStream` on a sidecar lock file) around the read-modify-write, or delete the temp file in a `catch`.

src/Configuration/SAF.Configuration.Secrets/Configuration/SecretConfigurationBuilderExtensions.cs
```diff
@@ -0,0 +1,54 @@
...
+    /// </summary>
+    internal static IConfigurationBuilder AddResolvedSecrets(
+        this IConfigurationBuilder builder,
+        HostSecretStoreAccessor? accessor,
+        Action<SecretStoreOptions>? configure = null,
+        Action<ISecretStoreBuilder>? configureProviders = null)
+    {
+        ArgumentNullException.ThrowIfNull(builder);
+
+        var innerSources = builder.Sources.ToList();
```
berndbr, line 51: **[Medium/Efficiency] Every pre-existing configuration source is built a second time, and each secret reference costs a full store round-trip with no caching — repeated on every reload.**

`builder.Sources.ToList()` snapshots the sources and hands them to `SecretResolvingConfigurationSource`, but they are **left in the outer builder as well**. `SecretResolvingConfigurationProvider`'s ctor then does `innerBuilder.Add(source)` for each and `innerBuilder.Build()`, so every source is materialized twice.

Concrete cost in the plugin-system path, where the sources are the plugin settings JSON plus its environment overlay:
- 2x `JsonConfigurationProvider` instances per settings file, 2x file read + parse at startup;
- 2x change-token registrations against the shared `PhysicalFileProvider` for the host's lifetime — note `PluginSystemHostContext` went out of its way to seed one shared file provider precisely to avoid duplicated watchers (*"All of those per-source providers would otherwise be undisposed, each holding a recursive FileSystemWatcher"*);
- every file edit fires both the outer root's reload **and** the inner root's reload -> `Reload()`, so `BuildData()` runs twice per change.

On top of that, resolution has no cache at any level. `BuildData()` calls `Resolve()` once per reference, and `FileSecretStore.GetSecretAsync` reads **and JSON-parses the entire store file** plus runs a CMS decrypt for each single lookup. A host with 20 secret references does 20 full file reads + 20 RSA-OAEP unwraps at startup, and repeats all of it on every appsettings change — all of it synchronously blocking via `.GetAwaiter().GetResult()` on the startup path.

Cheapest fixes: read the document once per `BuildData()` pass (or add a short-lived cache in `FileSecretStore` keyed on file mtime), and consider having the resolving source *replace* the captured sources in the outer builder rather than duplicating them.

src/Configuration/SAF.Configuration.Secrets/WindowsCredentialManager/WindowsCredentialManagerNativeApi.cs
```diff
@@ -0,0 +1,167 @@
...
+        {
+            Marshal.Copy(blob, 0, blobPtr, blob.Length);
+
+            var credential = new Credential
+            {
+                Flags = 0,
+                Type = CredTypeGeneric,
+                TargetName = targetName,
+                Comment = null,
+                CredentialBlobSize = (uint)blob.Length,
```
berndbr, line 71: **[Low-Medium] `CRED_MAX_CREDENTIAL_BLOB_SIZE` is not checked, so secrets over ~1280 characters fail with an opaque `Win32Exception`.**

`wincred.h` defines `CRED_MAX_CREDENTIAL_BLOB_SIZE` as `5 * 512` = **2560 bytes**. The blob here is UTF-16 (`Encoding.Unicode.GetBytes`), so the practical limit is ~1280 characters. `CredWriteW` rejects anything larger with `ERROR_INVALID_PARAMETER` (87).

**Failure scenario:** a product stores a PEM-encoded private key, a long-lived JWT, or a certificate passphrase bundle via `SetSecretAsync`. On Windows it fails with `Win32Exception: Writing the credential 'myapp/tls/key' failed.` and the native message "The parameter is incorrect" — which gives the caller no clue that a length limit was hit. The same payload works fine on the file provider, so the failure is silently platform-dependent.

Relatedly, `TargetName` is bounded by `CRED_MAX_GENERIC_TARGET_NAME_LENGTH` (32767 chars) and `UserName` — which this code sets to `targetName` — by `CRED_MAX_USERNAME_LENGTH` (513 chars), so a namespaced secret name longer than 513 characters also fails with the same opaque error.

A length pre-check with an explicit message ("the Windows Credential Manager limits a secret to 2560 bytes / 1280 UTF-16 characters; use the file provider for larger values") would turn a mystery into an actionable error. Worth a line in `docs/secret-store.md`'s provider table too, since it is a real functional difference between the two providers.

src/Configuration/SAF.Configuration.Secrets.Extensions.Tests/PluginSystemHostBuilderExtensionsTests.cs
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    [Fact]
+    public void AddSecretConfigurationResolution_ReturnsBuilderForChaining()
+    {
+        var services = new ServiceCollection();
+        var hostBuilder = Substitute.For<IPluginSystemHostBuilder>();
+        hostBuilder.Services.Returns(services);
+
+        var result = hostBuilder.AddSecretConfigurationResolution();
```
berndbr, line 82: **[Low-Medium/Test coverage] `AddSecretConfigurationResolution` — the feature's main entry point — has no behavioural test, and CI never exercises the non-Windows provider path.**

The only two tests for it assert that it returns the builder and throws on a null builder. Nothing invokes the callback it registers via `AddPluginConfigurationSource`, so none of the following is covered end-to-end:

- the bootstrap reader actually resolving a reference through the plugin config pipeline;
- the `HostSecretStoreAccessor` -> `HostSecretStoreAccessorInitializer` handover in the real plugin-system flow;
- composing `AddSecretStore` + `AddSecretConfigurationResolution` (the docs promise they "compose safely");
- the interaction with `PluginSystemHostContext.AddCustomPluginConfigurationSources`.

This is exactly the gap that lets the critical bootstrap/`ISecretProtector` bug ship: a single test that registers the protector the way the docs do and then resolves a `secret://` reference would fail immediately.

Compounding it, `.github/workflows/dotnet-core.yml` runs `dotnet build` / `dotnet test` **only** on `windows-latest`. So `AddDefaults()`'s `if (!OperatingSystem.IsWindows())` branch, `FileSecretStore` auto-selection, and the entire Linux code path this feature was written for are never executed in CI. The `authenticode_integration` job is the only ubuntu runner and it does not touch these projects.

Suggest adding (a) a test that drives `AddSecretConfigurationResolution` through a real `PluginSystemHostContext`, and (b) a linux leg to the `build_test_pack` job — or at minimum a test that forces the non-Windows registration path so it is covered on the Windows runner.

berndbr (commented):
Automated max-effort code review of the secret store feature (correctness + documentation consistency + security). 15 findings below, most severe first.

**Highest-impact items:** (1) the bootstrap DI container never sees the host-registered `ISecretProtector`, so `AddSecretConfigurationResolution` hard-fails at startup on Linux; (2) the atomic-replace write silently destroys the file ACL/mode that the docs delegate to the installer (verified empirically); (3) the stored document has no integrity protection at all; (4) an unresolvable reference silently becomes `null` instead of failing closed.

Build verified: `dotnet build` succeeds with 0 warnings.

docs/secret-store.md
berndbr: Fixed in 2f1865c: replaced the `PluginContractsSearchPattern` instruction with the actual mechanism (`IHostServiceForwarder` + automatic assembly sharing via the plugin system's default-context loading, see `PluginAssemblyLoadContext.IsLoadableInDefaultContext`), and added an explicit note against adding it to `PluginContractsSearchPattern` since that setting controls cross-plugin service export discovery instead.

src/Configuration/SAF.Configuration.Secrets/WindowsCredentialManager/WindowsCredentialManagerNativeApi.cs
berndbr: Fixed in a2824a9: `WindowsCredentialManagerSecretStore.SetSecretAsync` now validates against `CRED_MAX_CREDENTIAL_BLOB_SIZE` (2560 bytes UTF-16) and `CRED_MAX_USERNAME_LENGTH` (513 chars — the target name is also stored as the credential's user name) up front, throwing a diagnosable `ArgumentException` naming the limit instead of letting the bare `Win32Exception(87)` from `CredWriteW` surface.

src/Configuration/SAF.Configuration.Secrets/Configuration/SecretResolvingConfigurationProvider.cs
berndbr: Fixed in b9d6c30 — went with the "ideally" option: `Resolve()` now throws by default when the store returns `null` for a reference, naming both the reference and the configured provider, instead of writing `null` into `Data`:

```
The secret reference 'secret://app/db/missing' could not be resolved: no value named
'app/db/missing' was found by the 'auto' secret store provider. Set
ThrowOnUnresolvedReference = false to pass unresolved references through as null instead.
```

`SecretStoreOptions.RequireSecretReferences` — the option you correctly identified as documented-but-never-read — is gone; `ThrowOnUnresolvedReference` is its real, wired-up replacement (default `true`), kept as a distinctly-named opt-out for test/dev rather than reusing that name.

Also closes the inconsistency you flagged: an infrastructure failure and a missing secret are now both hard failures at startup, not one crash and one silent `null`. `SecretResolvingConfigurationTests.UnresolvableReference_Throws_ByDefault` pins the new behavior; `UnresolvableReference_BecomesNull_WhenThrowOnUnresolvedReferenceIsFalse` pins the explicit opt-out.

src/Configuration/SAF.Configuration.Secrets/Configuration/SecretResolvingConfigurationProvider.cs
berndbr: Fixed in b9d6c30, the way you suggested: `Reload()`'s body is now wrapped in a try/catch that logs a warning and keeps the previous `Data` instead of propagating.

Went with a blanket `catch (Exception ex)` rather than enumerating `JsonException`/`FormatException`/`CryptographicException`/`Win32Exception`/etc.: providers are pluggable (`ISecretStoreProvider` is a public extension point), so the set of exception types a store can throw isn't closed, and this boundary's job is specifically "never let a reload failure reach the watcher thread" regardless of cause — the same reasoning the plugin system's own `OnLoadException` guard applies to `FileConfigurationSource`.

Initial `Load()` is deliberately left unwrapped — an exception there (missing secret, or now also a genuine store failure) still propagates and fails startup, which is the desired fail-fast behavior from the fail-closed change (see the reply on the sibling "fail-open" comment). `Reload_ContainsProviderException_KeepsPreviousData_AndLogsWarning` pins the contained path; `InitialLoad_PropagatesProviderException` pins that `Load()` still fails hard.

src/Configuration/SAF.Configuration.Secrets/Configuration/SecretResolvingConfigurationProvider.cs
berndbr: Took the documentation mitigation in b9d6c30, per your own framing that this is "largely inherent to the `IConfiguration` model" — added to the "Security model" section (docs/secret-store.md) exactly where you pointed out it belongs:

> Resolved secrets are plaintext in the process's own configuration state. Once a `secret://` reference resolves, the value lives in `IConfiguration` like any other setting, so `IConfigurationRoot.GetDebugView()` prints it, attributed to the resolving provider. Use the `GetDebugView(processValue)` overload to mask values before logging or displaying a debug view of configuration that may include resolved secrets.

Deliberately did **not** adopt the "resolve lazily, only on `TryGet`" idea, and want to flag why rather than silently drop it: this PR's fail-closed default (a separate comment on this review) needs every reference resolved eagerly at `Load()`, specifically so a missing/misconfigured secret fails startup immediately instead of surfacing whenever something eventually reads that key — which could be much later, or never, for a rarely-touched config path. Lazy resolution would trade the fail-closed guarantee for reduced exposure; given a choice, fail-closed at startup won out here.

Agreed this doesn't remove plaintext from memory or make the `Array.Clear` scrubbing upstream any less cosmetic — not claiming otherwise, and not treating this as solved beyond the documented mitigation.

src/Configuration/SAF.Configuration.Secrets/FileStore/FileSecretStore.cs
berndbr: Fixed in 500e64b (superseding an interim attempt in 849f5c5 that still used a temp file — see below): `WriteDocumentAsync` no longer creates a second file at all. It opens the store file itself exclusively (`FileShare.None`, `FileMode.OpenOrCreate`) and writes the new JSON **in place** through that one handle, truncating with `stream.SetLength` if the new content is shorter than the old.

Since it's the same file/inode throughout, an existing file's ACL/mode is never replaced — permission preservation falls out for free instead of needing an explicit copy step. A brand-new file gets a restrictive Unix mode (`UserRead | UserWrite`) via `FileStreamOptions.UnixCreateMode`; on Windows it still gets whatever the parent directory grants (this provider still doesn't manage ACLs itself, per the documented "installer's responsibility" design — it just no longer *destroys* what the installer set).

Verified empirically, the same way this finding was: reproduced the `icacls` degradation you found (a protected, single-ACE file became 5 inherited ACEs after one `Move`), confirmed the rewritten in-place write leaves the ACL untouched on real Windows, and confirmed via a Linux container that a fresh file lands at `0600` and an existing file's mode survives an overwrite.

This also removes the "temp file left behind on failure, outside the destination's ACL" half of the finding — there's no temp file to leak. The no-temp-file design was actually forced by an independent, harder constraint (see the reply on the `SemaphoreSlim`/`.tmp` comment below): some deployment targets only permit writing an already-existing file, ruling out temp-file-and-replace entirely, not just as an ACL concern.

Trade-off worth flagging explicitly since it's a behavior change from "atomic replace": a crash mid-write can now leave the store file truncated/corrupt, since there's no temp file to fall back to. Accepted given the constraint above.

src/Configuration/SAF.Configuration.Secrets/FileStore/FileSecretStore.cs
berndbr: Fixed in 500e64b (following an interim attempt in 849f5c5 that added a sidecar `.lock` file — dropped, see below): every read and write now opens the **store file itself** exclusively (`FileShare.None`) for the duration of the operation, instead of relying on the in-process `SemaphoreSlim` alone. A second opener — this process or another (e.g. the installer from your failure scenario) — gets an `IOException` immediately and polls until the holder releases it, rather than racing a read-modify-write against it.

All three problems you listed are addressed by the same change:

1. **Lost update** — the installer and the service now genuinely can't be inside a read-modify-write at the same time; the exclusive hold on the file serializes them.
2. **Temp-file collision** — there's no temp file at all anymore, so nothing to collide on. (Not just randomized — removed outright.)
3. **Leftover temp file** — same reason, moot.

I initially reached for a randomized-temp-name + sidecar-lock-file design close to what you suggested, but an independent, harder constraint ruled out creating *any* second file: some deployment targets only permit writing an already-existing file, not creating new ones (temp file **or** lock file). So the lock had to be the target file's own exclusive hold, not a sidecar — which turned out to subsume the ACL-preservation fix too, since writing in place through the same handle never touches the file's permissions.

Verified against real disk (not just the mock) with two independent `FileSecretStore` instances — sharing no in-process state, standing in for two processes — concurrently writing distinct keys to the same file: no lost keys, no corruption.

Trade-off: no atomic replace anymore, so a crash mid-write can leave the file truncated/corrupt. Accepted given the no-new-files constraint above.

src/Configuration/SAF.Configuration.Secrets/Configuration/SecretResolvingConfigurationProvider.cs
berndbr: Fixed in 8e65c91 by removing the bootstrap container entirely, rather than teaching it about host registrations (your first option).

`PluginConfigurationSourceContext` now carries the host's `IServiceProvider` (`HostServices`), and `SecretResolvingConfigurationProvider` resolves `ISecretStore`/`SecretStoreOptions` from that provider directly — there is no separate minimal `ServiceCollection` anymore, so it is structurally impossible for a host-registered `ISecretProtector` (or any other host registration) to be invisible to it. This works because `PluginSystemHostContext` is only ever constructed from the factory in `HostApplicationBuilderExtensions.cs`, which already has the fully-built host `sp` in hand — the two-phase design's premise ("the DI container doesn't exist yet") didn't actually hold at that call site.

Added an end-to-end regression test through a real `HostApplicationBuilder`/`AddPluginSystem`/`IHost` (not a substituted builder) so this fails immediately if it regresses: `AddSecretConfigurationResolution_ResolvesSecretInPluginConfiguration_ThroughRealHost`.

Also traced the actual hosted-service startup ordering to confirm the old phase-2 accessor could never win its race against plugin startup (always lost), and checked a real downstream consumer of this package that had already hit this exact crash in practice and worked around it by duplicating provider registrations for the second pass — that workaround is no longer necessary against this branch.

src/Configuration/SAF.Configuration.Secrets.Extensions.Tests/PluginSystemHostBuilderExtensionsTests.cs
berndbr: Closed by 8e65c91 (test) + a2fbee2 (CI), on top of the Package A rewrite:

- Added `AddSecretConfigurationResolution_ResolvesSecretInPluginConfiguration_ThroughRealHost`: drives `AddPluginConfigurationSource` + `AddSecretConfigurationResolution` through a real `HostApplicationBuilder`/`AddPluginSystem`/`IHost`, then asserts the resolved value on the real `IPluginSystemHostContext.PluginConfiguration` — this is the real-`PluginSystemHostContext` test suggested here, and since it chains two custom plugin-configuration sources (an in-memory one plus the secret-resolution one), it also covers the `AddCustomPluginConfigurationSources` interaction you called out.
- Added a `test_linux` CI job (`ubuntu-latest`), so `AddDefaults()`'s non-Windows branch and the file-store path now run in CI.
- The bootstrap-reader / `HostSecretStoreAccessor` handover this asked about no longer exists to test — Package A deleted both classes, since this review (this comment included) showed the handover itself, not just its test coverage, was the bug.
- "`AddSecretStore` + `AddSecretConfigurationResolution` compose safely" is no longer only a docs claim: `AddSecretConfigurationResolution` now calls `AddSecretStore` internally, and every registration on that path (`TryAddSingleton` for the store, `TryAddEnumerable` for every provider) is idempotent — so calling both explicitly is safe by construction, not by convention.

Remaining test debt from the wider review (env-var mangling branch coverage, `Dispose()`, corrupt-store paths, etc.) is tracked separately and wasn't part of what this comment raised.

src/Configuration/SAF.Configuration.Secrets/Protection/PkcsSecretProtector.cs
berndbr: Not implemented — resolved as a documented, deliberate scope decision instead (345083f4).

The swap/replay/delete/bit-flip scenarios you listed are all correct, and the assessment that the protector stamp "adds no integrity" was already exactly right. After discussing it, write access to the secret store is being treated as **trusted**, not adversarial, for this stage of the feature — whoever provisions the store (an installer, an admin tool) is assumed authorized. Given that, `SignedCms`/HMAC integrity isn't being added right now; it would protect against a threat that's explicitly out of scope here.

What did land is your own suggested minimum: `docs/secret-store.md`'s "Security model" section now states this explicitly, rather than leaving it implied. New bullet:

> **The writer is trusted.** Provisioning (whoever calls `SetSecretAsync` — an installer, an admin tool) is assumed authorized, not adversarial. In particular, `PkcsSecretProtector`'s PKCS#7/CMS envelope encrypts each value (AES-256-CBC content key wrapped with RSA-OAEP) but does not authenticate it: `Protect` only needs the recipient certificate's *public* key, so anyone who can write the store file could swap, replay, or forge entries without needing the private key. This is an accepted trade-off given the trusted-writer assumption, not an oversight — if your deployment cannot guarantee that, do not rely on this store's write path as a tamper-resistant boundary.

The "Protector identity is stamped" bullet is also now qualified as a misconfiguration safeguard, not tamper detection, for the same reason.

If your deployment can't guarantee a trusted writer, this is the place to push back — happy to pick `SignedCms`/HMAC-over-the-document back up as a follow-on if the threat model needs it.

docs/secret-store.md
berndbr: Fixed in 35e2fc0. Both examples now carry the logical name only — `secret://opcua/conn-1/user` / `secret://opcua/conn-1/password` — and the env-var bullet follows suit (`secret://opcua/conn-1/password` → `SECRET__opcua__conn_1__password`).

Since the docs contradicting themselves is what made this possible, the convention is now stated outright rather than only implied, in a callout directly under the JSON block:

> **A reference does not repeat the namespace.** It carries the *logical* name only; the store prepends the configured `Namespace` to form the physical key. With `o.Namespace = "myapp"`, `secret://opcua/conn-1/password` is looked up as `myapp/opcua/conn-1/password`. Writing `secret://myapp/opcua/conn-1/password` instead would look up `myapp/myapp/opcua/conn-1/password` and fail to resolve.

One detail in your failure scenario has changed since the review: the silent `null` is gone. Unresolvable references throw by default now (`ThrowOnUnresolvedReference`, b9d6c30), so copying the old example would have failed loudly at startup rather than binding `null`. The doc fix is still needed — but the diagnosis path is much shorter than when you filed this.

The tests could not have caught it, which felt like the more important gap: every fake provider in `SecretResolvingConfigurationTests` ignores `Namespace` entirely, so both candidate designs pass identically. Two new tests there now pin the convention end-to-end through a *real* store (`AddFile` over a mock filesystem):

- `Reference_CarriesTheLogicalNameOnly_AndTheStorePrependsTheNamespace` — asserts the key actually persisted is `myapp/opcua/conn-1/password`
- `Reference_ThatRepeatsTheNamespace_DoesNotResolve` — asserts the doubled form fails closed

Noticed while fixing this, and deliberately left alone here: the environment-variable name is derived from the reference name *before* the namespace is prepended, so it contains no namespace at all. Two hosts sharing an environment with different `Namespace` values collide on the same `SECRET__*` variable. Changing that is the same compatibility break as the mangling issue you raised on `BuildEnvironmentVariableName`, so the two will be decided together rather than piecemeal.

src/Configuration/SAF.Configuration.Secrets.Contracts/FileSecretStoreOptions.cs
berndbr: Removed, in 45317e1 — and `SecretStoreOptions.Scope` and the `SecretScope` enum with it.

Taking the "either implement it, or remove it" fork you offered. File permissions are a deliberate design decision on this feature: they are the installer's/deployment's responsibility, and the provider reads and writes file *content* only. So `ReaderPrincipal` had no implementation path to grow into. Neither did `Scope`, which only ever produced a log warning saying it could not be honoured — and once `ReaderPrincipal` went, `SecretScope` had no consumer left at all, so the enum went too. Your framing was the deciding argument: a security option that ships in IntelliSense promising exclusive read access, and does nothing, is worse than one that is absent.

Two knock-on effects worth naming:

1. That warning was `WindowsCredentialManagerSecretStore`'s only use of its logger, so the `ILogger` dependency went with it. The class is `internal`, so no public API break.
2. Removing `ReaderPrincipal` also removes the last hint **in code** about *which* principal an installer is expected to grant access to. That contract now lives only in the docs, so `docs/secret-store.md`'s "File permissions are the installer's responsibility" bullet is no longer just a statement of policy — it carries concrete commands against the documented default paths:

```powershell
icacls "$env:ProgramData\myapp\secrets.json" /inheritance:r /grant "NT SERVICE\MyApp:(R)" /grant "BUILTIN\Administrators:(F)"
```

```bash
chown myapp:myapp /var/lib/myapp/secrets.json && chmod 600 /var/lib/myapp/secrets.json
```

The contradiction you spotted against the "does **not** set ACLs" line resolves itself — that statement is now the only claim the docs make on the subject.

The two round-trip tests went with the property, as did `SecretStoreOptionsTests`' dead `Scope` and `ReaderPrincipal` default assertions.

src/Configuration/SAF.Configuration.Secrets/Configuration/SecretResolvingConfigurationProvider.cs
berndbr: Fixed in 4642eb2, taking your secondary concern as the primary one.

**`AllowEnvironmentOverride` now defaults to `false`.** Your framing of it — anyone who can influence the environment block can substitute any secret without touching the Credential Manager or the encrypted file, defeating the point of the store — is the deciding argument, and it isn't something a doc note can offset while the behaviour is on by default. CI and dev opt in with one line.

**The variable name now includes the namespace.** Not in your comment, but found while fixing the namespace doubling in the docs (your other comment): the name was derived from `reference.Name` *before* the namespace was prepended, so it contained no namespace at all. Two hosts sharing an environment under different `Namespace` values collided on one variable — exactly the collision the namespace exists to prevent. It now derives from the physical store key via `SecretTargetName.Build`, so `myapp` and `other` get `SECRET__myapp__…` and `SECRET__other__…`. Side effect worth knowing: that key is lower-cased, so the derived name is too, which matters on Linux where env lookup is case-sensitive. Documented.

**`EnvironmentVariablePrefix` is validated** — anything but letters, digits and underscores produces a variable name that can never be set, so it now throws naming the option instead of silently never matching. **An applied override is logged at `Debug`** with the reference and variable name, never the value.

**On the mangling: deliberately not made injective.** This one went the other way, and I want to be explicit about it rather than quietly leave it. Your table is correct and the collisions are real. But the options were hex-escaping (`SECRET__myapp__conn_2D1__password`) or restricting the legal charset for secret names, and both make a name that an operator types by hand materially worse in order to fix a case that requires two secrets differing *only* in their non-alphanumeric characters. With the override now off by default, that case is reachable only where someone has deliberately enabled it.

So it is documented instead of fixed, in a new "Environment overrides" section that spells out the collisions with your own examples, states that an override wins over a correctly provisioned keystore entry unconditionally and never passes through the store's at-rest protection, and says plainly that enabling the option widens the trust boundary to everyone who can set the process environment. The `Security model` section gained a matching bullet, since that is where a security reader looks first.

If you think the collision warrants closing properly despite the ergonomics, say so and I'll take the charset-restriction route — it is the cheaper of the two and this is the moment to do it, before anything depends on the current names.

Tests: the override tests no longer lean on the default (which would now assert nothing), and cover the opt-in, the default-off path, namespace isolation, and the invalid prefix. Worth flagging that the N3 regression test had to be inverted — it proved options reach the resolver by setting `AllowEnvironmentOverride = false`, which is now the default and therefore proves nothing; it sets `true` and asserts the override wins instead. 130 passed / 2 skipped.

src/Configuration/SAF.Configuration.Secrets/CompositeSecretStore.cs
berndbr: Fixed in 8e27c3e — `LazyThreadSafetyMode.PublicationOnly`, so a successful selection is still cached and a failed one is retried on the next call.

One correction to the scenario, because it changes who this protects rather than whether it is worth fixing: with the two built-in providers, availability is not transient. `WindowsCredentialManagerSecretStore.IsAvailable` is `OperatingSystem.IsWindows()`, and `FileSecretStore.IsAvailable` tests `_protector is not null` on a field injected at construction — a retry returns the same answer either way. The latch bites the extension point the docs explicitly invite: a custom `ISecretStoreProvider` over a remote vault, whose availability *is* a runtime fact. That is the case that could previously never recover without a restart, and it is the one now covered by `FailedSelection_IsNotCached_SoAProviderThatBecomesAvailableIsUsed`.

`PublicationOnly` lets concurrent first callers each run `SelectProvider`, with the losing results discarded. That is safe here — it is a query over an already-materialized provider list — at the cost of the "Auto-selected secret store provider" line possibly appearing twice under a startup race. Accepted rather than papered over with an `Interlocked` dance.

**On the other half of this, which your comment did not raise but the fix forces a position on:** selection stays *one-shot*. `GetSecretAsync` delegating to a single provider is not an oversight to be fixed into a fallback chain, and the docs were what made it look like one — the custom-provider sample literally labelled the second provider `// fallback`. That wording is gone, and the selection section now states the semantics outright:

> **One provider answers, and it is not a fallback chain.** Selection happens once; the chosen provider serves every read *and* every write for the process lifetime. A name it does not hold is absent — the next registered provider is never consulted for it.

The reasoning, in case you would have gone the other way: a per-lookup chain makes "which backend answered?" unanswerable, which for a *secret* store is an audit loss; it lets a stale value in a lower-priority store shadow the authoritative one; it splits reads from writes unless `SetSecretAsync` also walks the chain, which then contradicts itself; and it turns the fail-closed diagnostic added in b9d6c30 from "this reference did not resolve" into "four backends were asked and none had it". `Miss_DoesNotFallBackToTheNextProvider` now pins that, so the semantics cannot drift back silently.

If you read the priority-order wording as a promise of real fallback and want it implemented rather than corrected, say so — but it is a behaviour change with the write-path question attached, so I would rather settle it here than assume.

src/Configuration/SAF.Configuration.Secrets/Configuration/SecretConfigurationBuilderExtensions.cs
berndbr: Fixed in fbfd75d — all four parts, though not by the route you suggested, and the change reaches outside this feature. Flagging that up front so you can push back on it.

**Replacing the captured sources in the outer builder does not work.** I tried it first. `Sources.Clear()` is fatal on `ConfigurationManager` (the usual host builder type), because its `Build()` returns `this` — clearing afterwards empties the very root you just captured. Swapping the entries in place instead trips `List<T>`'s version check inside the outer `Build()`'s own `foreach`. Doing it at add-time works mechanically but turns `builder.Sources[i]` into wrappers, which breaks the `is FileConfigurationSource` loop in `PluginSystemHostContext` that applies `OnLoadException` to custom sources.

The underlying reason is worth stating: a provider that has to read the *composed* configuration cannot be a peer in the chain that composes it. The composition does not exist until every source is built.

**So the fix adds a hook.** `PluginConfigurationSourceContext.DecorateConfigurationRoot(Func<IConfigurationRoot, IConfigurationRoot>)`, applied after every source is built, and the plugin integration chains the root that already exists (`AddConfiguration(root, shouldDisposeConfiguration: true)`) instead of rebuilding it. Each settings file is now parsed once and watched by one watcher; a test asserts `BuildCount == 1` through a real host. That is new public API on `SAF.PluginSystem.Hosting` — a released package, for a problem the secret store happened to surface. I kept it general (it is about reading composed configuration, not about secrets) and documented it in `docs/plugin-system.md`, but if you would rather this stayed inside the secrets package at the cost of leaving the double materialization in place, say so and I will revert to that.

**It also fixes the ordering trap, which is the more serious half.** The snapshot meant a source added by a *later* `AddPluginConfigurationSource` callback was never resolved *and* won `TryGet` — the plugin received the literal `secret://…` token as its credential. Under the decorator, registration order stops mattering entirely. For the standalone `AddResolvedSecrets(IConfigurationBuilder, …)`, which has no control over the root, the sources are now read at `Build()` time and a reference it cannot override throws at startup naming the keys, rather than passing the token through. That last one is a behaviour change: a build that silently produced a wrong credential now fails.

**One thing I did not expect to find.** Muting the chained root's reload token is not cosmetic. Reload-token callbacks run last-registered-first, so forwarding it would have notified consumers *before* the resolver caught up — and a reference introduced by that very reload would have read as its literal token in the window. The resolver now raises the single notification, after re-resolving. `ChainedRoot_ResolvesAReferenceThatOnlyAppearsOnReload` fails without it.

**Caching and the blocking call**, per your last paragraph:

- `BuildData` resolves once per *distinct reference name*, not once per configuration key.
- `FileSecretStore` keeps the parsed document between reads, revalidated against last-write-time **and** size, dropped whenever it writes. Worth knowing that this is a weak validator: two distinct writes of identical size landing on the same timestamp tick would serve a stale document. Not realistically reachable on NTFS/ext4, but it is not a proof. Only the encrypted document is cached — never a decrypted value, so nothing sits in memory in the clear that did not already.
- The `GetAwaiter().GetResult()` per reference is now one wait for the whole load, dispatched through `Task.Run` so it cannot deadlock on a captured `SynchronizationContext`, and bounded by a new `SecretStoreOptions.ResolveTimeout` (30s, `Timeout.InfiniteTimeSpan` to opt out). There is no `CancellationToken` to plumb through — `IConfigurationProvider.Load()` has none — so the real defect was the *unbounded* wait: an unreachable vault used to hang startup with no diagnostic and now fails with a named timeout.

Your 20-reference example goes from 20 file reads + 20 unwraps per load to one read and one unwrap per distinct secret.

150 passed / 2 skipped in the secrets suite, 9/9 extensions, 159/159 plugin system, 0 new warnings. I verified the cache test actually discriminates by disabling the cache and watching it fail.

